Issue #10, opened 2018-03-23
Run Cosmas inside Docker container


dbarashev:
We can run Cosmas with `gradle run`, now let's take a step further and package it into docker container.

We need a Dockerfile which would package Cosmas distribution (server component in particular) into docker image based e.g. on [openjdk:8-slim](https://hub.docker.com/_/openjdk/). Please notice that we don't want to run `gradle` in the container, we want to run already compiled application. `gradle distZip` will build a ZIP file with a ready to run app and will put it into `build/distributions` directory.

SashaOrlova:
did I understand right, that Dokerfile need unzip file and run it?

dbarashev:
Well, you don't want to run the application in dockerfile, if you meant exactly [RUN instruction](https://docs.docker.com/engine/reference/builder/#run), because it runs when you build the image, not when you run the container. But yes, if you unzip and make an entrypoint from the launcher script, it should do the trick.

PR #12, opened 2018-03-26
run inside Docker container

SashaOrlova:
#10


Dockerfile
```diff
@@ -1,7 +1,8 @@
-FROM openjdk:8-jdk
+FROM openjdk:8-slim
 
-ENV FILE_NAME distZip
+ENV DIR_NAME Cosmas-1.0-SNAPSHOT
```
dbarashev: Cool, this looks way better. The remaining issues:
1. The name of this variable is misleading. It is not a directory name, although it appears in the directory name by coincidence. It is rather cosmas version or something like that
2. It is needed at build time only (thsi value in the runtime is pretty useless), yet it is impossible to pass the value of this environment from the outside during build time. Instruction [`ARG`](https://docs.docker.com/engine/reference/builder/#arg) is designed for configuring dockerfile and build time.

dbarashev (approved):
LGTM. You may want to supply the default value of ARG to avoid typing it every time.

Issue #11, opened 2018-03-26
Use Google Cloud Storage for storing files


dbarashev:
Let's add another implementation of CosmasService which uses a real storage option for our file versions: Google Cloud Storage.

> Google Cloud Storage is a binary object storage which supports object versioning. It mostly makes sense when application is running in Google Cloud Platform datacenter, but it can also be accessed from anywhere in the world. You can find good overview and tutorials [in the docs](https://cloud.google.com/storage/docs/)

You are mostly interested in reading and writing objects and not interested in managing buckets. I'll create a test bucket for you and in production we'll use an already created bucket (so please pass the bucket name in the command line args).

For the first step we may ignore versioning at all and just write and read the latest version of the object, as if there is just one version.

You will also need a credentials file which I will send separately.

Also, please do not throw away in-memory implementation. We'll need both of them.

PR #13, opened 2018-04-02
Google Cloud storage

iisuslik43:
#11


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private val storage: Storage = StorageOptions.getDefaultInstance().service
+
+    override fun createVersion(request: CosmasProto.CreateVersionRequest,
+                               responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
+        println("Get request for create new version of file № ${request.fileId}")
+        storage.create(
+                BlobInfo.newBuilder(bucketName, request.fileId)
+                        // Modify access list to allow all users with link to read file
+                        .setAcl(ArrayList(Arrays.asList(Acl.of(User.ofAllUsers(), Acl.Role.READER))))
```
dbarashev: Please remove this. Do you really want anyone with the link to be able to read your personal data?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,72 @@
...
+                .newBuilder()
+                .build()
+        responseObserver.onNext(response)
+        responseObserver.onCompleted()
+    }
+
+    override fun getVersion(request: CosmasProto.GetVersionRequest,
+                            responseObserver: StreamObserver<CosmasProto.GetVersionResponse>) {
+        println("Get request for version ${request.version} file № ${request.fileId}")
+        val blob = storage.get(BlobInfo.newBuilder(bucketName, request.fileId).build().blobId)
```
dbarashev: It is pretty important to be able to understand from the context the scope of variables and distinguish local vars/function arguments from instance or class members. It can be achieved by accessing instance members with `this.` or by prefixing them with `my`, `_` or whatever else. It seems that Kotlin world prefers `this.`, so shall we. Please do use `this.` when accessing instance variables: `this.storage`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,72 @@
...
+        val response = CosmasProto.GetVersionResponse.newBuilder()
+        if (blob != null) {
+            response.file = ByteString.copyFrom(blob.getContent())
+        } else {
+            val requestStatus = Status.NOT_FOUND.withDescription("There is no such file in storage")
+            println("This request is incorrect: " + requestStatus.description)
+            responseObserver.onError(StatusException(requestStatus))
```
dbarashev: You may want to return at this point. There is no point in subsequent call of `responseObserver.onNext`

Another option, which is even better in spite of exception handling is to send `onNext` and `onCompleted` only from successful branch.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    override fun getVersion(request: CosmasProto.GetVersionRequest,
+                            responseObserver: StreamObserver<CosmasProto.GetVersionResponse>) {
+        println("Get request for version ${request.version} file № ${request.fileId}")
+        val blob = storage.get(BlobInfo.newBuilder(bucketName, request.fileId).build().blobId)
+        val response = CosmasProto.GetVersionResponse.newBuilder()
+        if (blob != null) {
+            response.file = ByteString.copyFrom(blob.getContent())
+        } else {
+            val requestStatus = Status.NOT_FOUND.withDescription("There is no such file in storage")
```
dbarashev: You may also want to handle `StorageException` and return appropriate statuses

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -64,7 +67,7 @@ class CosmasService : CosmasGrpc.CosmasImplBase() {
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
-        println("Get request for create new version")
+        println("Get request for create new version of file № ${request.fileId}")
```
dbarashev: Using `№` is quite uncommon for indicating id or number. Programmers are more likely to see and understand `#`, so let's use it.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
```diff
@@ -50,12 +51,17 @@ class CosmasServer(port: Int) {
 fun main(args: Array<String>) {
     val arg = CosmasServerArgs(ArgParser(args))
     println("Try to bind in port ${arg.port}")
-    val server = CosmasServer(arg.port)
+    val server =
+            if (arg.bucket != "")
+                CosmasServer(arg.port, CosmasGoogleCloudService(arg.bucket))
+            else
+                CosmasServer(arg.port, CosmasInMemoryService())
```
dbarashev: Please always use curly braces in the control flow statements. The only exception is when you write the whole statement on a single line: `if (condition) then doThis() else doThat()`

The reason is that code without curly braces is way more error prone. Should you insert a single logging line above this line, and your code turns into a pumpkin.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
```diff
@@ -50,12 +51,17 @@ class CosmasServer(port: Int) {
...
+                CosmasServer(arg.port, CosmasGoogleCloudService(arg.bucket))
+            else
+                CosmasServer(arg.port, CosmasInMemoryService())
     println("Start working in port ${arg.port}")
     server.start()
     server.blockUntilShutDown()
 }
 
 class CosmasServerArgs(parser: ArgParser) {
     val port: Int by parser.storing("--port", help = "choose port") { toInt() }.default { 50051 }
```
dbarashev: nitpicking: _choose_ is not so informative. Of course I have to choose because otherwise I can't run your server. _What is the port_ I have to choose? What's the meaning? Server port? Some other port?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
```diff
@@ -50,12 +51,17 @@ class CosmasServer(port: Int) {
...
+    val bucket: String by parser.storing("--bucket", help = "choose bucket")
```
dbarashev: The same about bucket. `help="GCS bucket name for file storage" ` gives at least an idea that it is not an ice bucket

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    companion object {
+        fun getServiceForTests(): CosmasGoogleCloudService {
+            return CosmasGoogleCloudService("papeeria-interns-cosmas",
+                    LocalStorageHelper.getOptions().service)
+        }
+    }
+
+    private constructor(bucketName: String, testStorage: Storage) : this(bucketName) {
```
dbarashev: Please move this upwards, after member fields declaration. Constructors are important guys and deserve being placed the first.

Another option is to add storage argument to the primary constructor and provide a default value (the real storage)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Special class that can work with requests from CosmasClient
+ * This realization stores files in Google Cloud Storage.
+ *
+ * @author Aleksandr Fedotov (iisuslik43)
+ */
+class CosmasGoogleCloudService(private val bucketName: String) : CosmasGrpc.CosmasImplBase() {
+
+    private var storage: Storage = StorageOptions.getDefaultInstance().service
```
dbarashev: This can be `val`, I believe.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,84 @@
...
+            println("This request is incorrect: " + requestStatus.description)
+            responseObserver.onError(StatusException(requestStatus))
+        }
+        responseObserver.onNext(response.build())
+        responseObserver.onCompleted()
+    }
+
+    fun deleteFile(fileId: String) {
+        println("Delete file № $fileId")
+        storage.delete(BlobInfo.newBuilder(bucketName, fileId).build().blobId)
```
dbarashev: Catch StorageException.

build.gradle
```diff
@@ -43,6 +43,9 @@ dependencies {
     compile "io.grpc:grpc-netty:1.+"
     compile "io.grpc:grpc-testing:1.+"
     compile "com.xenomachina:kotlin-argparser:2.0.4"
+    compile 'com.google.cloud:google-cloud-storage:1.14.0'
```
dbarashev: replace the last numbers with `+` ? Things change so frequently these days...

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: No, it works only with `var` because I change this field in the second constructor

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: Ok, but I like private constructor

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

build.gradle
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
dbarashev: That's another reason to move it into primary constructor args.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
dbarashev: Private constructors are nice when they have some good reasons behind. Do not allow anyone to create instances of this class (and restrict to a set of predefined instances) is a good reason. But here you actually do need to parameterize this class with different `Storage` implementations. If you start writing tests for the cases when `Storage` throws exceptions (and you normally want such tests, because they do throw exceptions) you will find our that throwing exception from the real or in-memory implementation is not that easy. The easiest way is using mocks. And the only way to pass mock implementation from tests is to allow for passing `Storage` instance.

In general, usually if your code depends on some non-trivial component, it is a good idea to parameterize your code with instance of that component interface, rather than access the component instance from the code itself. Parameterizing makes your code more test-friendly and more change-friendly: you explicitly declare the interface you depend on instead of hiding such interface in the class internals.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,97 @@
...
+    fun deleteFile(fileId: String) {
+        println("Delete file # $fileId")
+        try {
+            this.storage.delete(BlobInfo.newBuilder(this.bucketName, fileId).build().blobId)
+        } catch (e: StorageException) {
+            println("Deleting file failed: ${e.message}")
+        }
+    }
+
+    companion object {
```
dbarashev: ... and if you move this stuff to the testing code you will actually have better separation between production and testing. You only need an object implementing `Storage` interface in this service implementation, you don't care if it comes from tests or from the real GCS

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServer.kt
```diff
@@ -39,23 +40,29 @@ class CosmasServer(port: Int) {
...
+                CosmasServer(arg.port, CosmasInMemoryService())
+            }
     println("Start working in port ${arg.port}")
     server.start()
     server.blockUntilShutDown()
 }
 
 class CosmasServerArgs(parser: ArgParser) {
-    val port: Int by parser.storing("--port", help = "choose port") { toInt() }.default { 50051 }
+    val port: Int by parser.storing("--port", help = "choose port that server will hear") { toInt() }.default { 50051 }
```
dbarashev: programmers usually use "listen to" when talking about ports

iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: Ok, thanks for your explanations, I deleted this private constructor

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -0,0 +1,85 @@
...
+        println("Get request for version ${request.version} file # ${request.fileId}")
+        val blob: Blob? = try {
+            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.fileId).build().blobId)
+        } catch (e: StorageException) {
+            println("StorageException happened: ${e.message}")
+            responseObserver.onError(e)
+            return
+        }
+        val response = CosmasProto.GetVersionResponse.newBuilder()
+        if (blob != null) {
```
dbarashev: Mostly aesthetic comment: if you swap the branches of this if, and process `if (blob == null ) { ... return }` branch first, you'll have two cases when you return error next to each other and final successful case at the end. This will make the successful case better visible.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @author Aleksandr Fedotov (iisuslik43)
+ */
+class CosmasGoogleCloudServiceTest {
+
+    private var service = getServiceForTests()
+
+    @Before
+    fun testInitialization() {
+        println()
```
dbarashev, line 35: Do you really need it ;) ?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test
+    fun addFileAndGetFile() {
+        addFileToService("file", "43")
+        val file = getFileFromService(0, "43")
+        assertFalse(file.isEmpty)
+        assertEquals("file", file.toStringUtf8())
+        service.deleteFile("43")
```
dbarashev: Use `this.` as well

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServiceTest.kt
```diff
@@ -18,19 +18,21 @@ import com.google.protobuf.ByteString
...
 
 /**
  * This is some tests for CosmasService class
  * @author Aleksandr Fedotov (iisuslik43)
  */
-class CosmasServiceTest {
+class CosmasInMemoryServiceTest {
 
-    private var service = CosmasService()
+    private var service = CosmasInMemoryService()
```
dbarashev: You can probably use `lateinit` instead of excessive initialization

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: Yes, i want to split tests log with this to make it readable

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasServiceTest.kt
iisuslik43: done

dbarashev (approved):
LGTM.
Good job! You can merge it now.

Issue #15, opened 2018-04-04
Enhance Cosmas service with a method to send edit patches


dbarashev:
Cosmas GRPC service currently provides methods to save new file version and get some specific version. We also want to record what happens between versions: who changes what at what time. For that purpose we need a method which takes this information and stores it in-memory or persistently.

Let's start with in-memory implementation. You have to add method in the service declaration in  the proto file and implement it in the in-memory impl 

User who makes the changes is identified by short string id.
File is identified by short string id a well.
The patch itself is a string, but it can be pretty large. For the purpose of this task the contents of that string doesn't matter
Modification timestamp is 64-bit integer.

PR #16, opened 2018-04-05
Cosmas service with a method to send edit patches

SashaOrlova:
#15

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
@@ -29,6 +29,14 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
 
     private val files = mutableMapOf<String, MutableList<ByteString>>()
 
+    private val versions = mutableMapOf<String, MutableList<storedStructure>>()
+
+    private class storedStructure(user1: String, text: String, time: Long) {
+        val user = user1
+        val patch = text
+        val timeStamp = time
+    }
+
     override fun getVersion(request: CosmasProto.GetVersionRequest,
                             responseObserver: StreamObserver<CosmasProto.GetVersionResponse>) {
         println("Get request for version ${request.version} file # ${request.fileId}")
@@ -76,11 +84,65 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
         responseObserver.onCompleted()
     }
 
+    private fun verifyGetHistoryOfVersionRequest(request: CosmasProto.GetHistoryOfVersionRequest): Status {
+        var status: Status = Status.OK
+        val fileVersions = this.versions[request.fileId]
+        when {
+            fileVersions == null ->
+                status = Status.INVALID_ARGUMENT.withDescription(
+                        "There is no file in storage with file id ${request.fileId}")
+            request.version >= fileVersions.size ->
+                status = Status.OUT_OF_RANGE.withDescription(
+                        "In storage this file has ${fileVersions.size} versions, " +
+                                "but you ask for version ${request.version}")
+            request.version < 0 ->
+                status = Status.OUT_OF_RANGE.withDescription(
+                        "You ask for version ${request.version} that is negative")
+        }
+        return status
+    }
+
     private fun addNewVersion(request: CosmasProto.CreateVersionRequest) {
         synchronized(this.files) {
             val fileVersions = this.files[request.fileId] ?: mutableListOf()
             fileVersions.add(request.file)
             this.files[request.fileId] = fileVersions
+        }
+    }
+
+    override fun createHistoryOfVersion(request: CosmasProto.CreateHistoryOfVersionRequest,
+                                        responseObserver: StreamObserver<CosmasProto.CreateHistoryOfVersionResponse>) {
+        println("Get request for create new history of file # ${request.fileId}")
+        synchronized(this.versions) {
+            val historyOfFile = versions[request.fileId] ?: mutableListOf()
+            historyOfFile.add(storedStructure(request.user, request.patch, request.timeStamp))
+            versions[request.fileId] = historyOfFile
+        }
+        val response: CosmasProto.CreateHistoryOfVersionResponse = CosmasProto.CreateHistoryOfVersionResponse
+                .newBuilder()
+                .build()
+        responseObserver.onNext(response)
+        responseObserver.onCompleted()
+    }
+
+    override fun getHistoryOfVersion(request: CosmasProto.GetHistoryOfVersionRequest,
+                                      responseObserver: StreamObserver<CosmasProto.GetHistoryOfVersionResponse>) {
+        println("Get request for history ${request.version} file # ${request.fileId}")
+        val response = CosmasProto.GetHistoryOfVersionResponse.newBuilder()
+        synchronized(this.versions) {
+            val fileHistoryOfVersions = this.versions[request.fileId]
+            val requestStatus = verifyGetHistoryOfVersionRequest(request)
+            if (requestStatus.isOk) {
+                val history = fileHistoryOfVersions?.get(request.version)
+                response.user = history?.user
+                response.patch = history?.patch
+                response.timeStamp = history?.timeStamp ?: 0
+                responseObserver.onNext(response.build())
+                responseObserver.onCompleted()
+            } else {
+                println("This request is incorrect: " + requestStatus.description)
+                responseObserver.onError(StatusException(requestStatus))
+            }
         }
     }
 }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
@@ -29,13 +29,9 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
 
     private val files = mutableMapOf<String, MutableList<ByteString>>()
 
-    private val versions = mutableMapOf<String, MutableList<storedStructure>>()
+    private val patches = mutableMapOf<String, MutableList<Patch>>()
 
-    private class storedStructure(user1: String, text: String, time: Long) {
-        val user = user1
-        val patch = text
-        val timeStamp = time
-    }
+    data class Patch(val user: String, val text: String, val timeStamp: Long)
 
     override fun getVersion(request: CosmasProto.GetVersionRequest,
                             responseObserver: StreamObserver<CosmasProto.GetVersionResponse>) {
@@ -84,24 +80,6 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
         responseObserver.onCompleted()
     }
 
-    private fun verifyGetHistoryOfVersionRequest(request: CosmasProto.GetHistoryOfVersionRequest): Status {
-        var status: Status = Status.OK
-        val fileVersions = this.versions[request.fileId]
-        when {
-            fileVersions == null ->
-                status = Status.INVALID_ARGUMENT.withDescription(
-                        "There is no file in storage with file id ${request.fileId}")
-            request.version >= fileVersions.size ->
-                status = Status.OUT_OF_RANGE.withDescription(
-                        "In storage this file has ${fileVersions.size} versions, " +
-                                "but you ask for version ${request.version}")
-            request.version < 0 ->
-                status = Status.OUT_OF_RANGE.withDescription(
-                        "You ask for version ${request.version} that is negative")
-        }
-        return status
-    }
-
     private fun addNewVersion(request: CosmasProto.CreateVersionRequest) {
         synchronized(this.files) {
             val fileVersions = this.files[request.fileId] ?: mutableListOf()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -29,6 +29,14 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
 
     private val files = mutableMapOf<String, MutableList<ByteString>>()
 
+    private val versions = mutableMapOf<String, MutableList<storedStructure>>()
+
+    private class storedStructure(user1: String, text: String, time: Long) {
```
dbarashev: * Please do follow standard naming conventions. ClassNamesAreCapitalized.
* Please do not use names like "user1". There must be really good reasons for that.
* Please do not use names like "someStuff" (which is nearly equivalent to "storedStructure"). Stored structure is pretty much anything, while for this thing there must be good name which reflects its semantics.
* For such simple structures  Irecommend using  Kotlin's [data classes](https://kotlinlang.org/docs/reference/data-classes.html)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -29,6 +29,14 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
 
     private val files = mutableMapOf<String, MutableList<ByteString>>()
 
+    private val versions = mutableMapOf<String, MutableList<storedStructure>>()
```
dbarashev: What you are going to keep here is not versions. Versions are stored in the map above. Single edit is not a version.
I would call these things "patches" unless you have a better name. You have a base version, you apply patches.

src/main/proto/cosmas.proto
```diff
@@ -37,9 +37,34 @@ message CreateVersionRequest {
...
+    string patch = 3;
+    int64 timeStamp = 4;
+}
+
+message CreateHistoryOfVersionResponse {
+}
+
+message GetHistoryOfVersionRequest {
+    string fileId = 1;
+    int32 version = 2;
```
dbarashev: What's the meaning of this "version" field?

src/main/proto/cosmas.proto
```diff
@@ -37,9 +37,34 @@ message CreateVersionRequest {
 message CreateVersionResponse {
 }
 
+message CreateHistoryOfVersionRequest {
```
dbarashev: When you type a few words in the editor and send a small patch to Cosmas, you do not create a new version, because version is the whole file text, not a single patch. You do not create a new history either, because history by definition is an (order) sequence of some other actions. 
What you do when you send a small patch you, well, create a new patch. 

Let's name the things appropriately. When I read "CreateHistoryOfVersion" I think that this method creates a history of versions. When I learn that actually it adds a patch, I am surprised and my WTF/second rate grows. When you write code your goal (besides writing  code that works of course) is to minimize that WTF/second rate for those who will read your code.

src/main/proto/cosmas.proto
```diff
@@ -37,9 +37,34 @@ message CreateVersionRequest {
...
+    string user = 1;
+    string fileId = 2;
+    string patch = 3;
+    int64 timeStamp = 4;
+}
+
+message CreateHistoryOfVersionResponse {
+}
+
+message GetHistoryOfVersionRequest {
```
dbarashev: There were no requirement to add get request in the issue (it talks about sending patches only), but since you added it, what in your opinion is the scenario of using the patch history? How will it look like from user perspective and how will it be translated to RPC requests? Do you probably need to change the request and response to fit the access pattern better?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -76,11 +84,65 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
         responseObserver.onCompleted()
     }
 
+    private fun verifyGetHistoryOfVersionRequest(request: CosmasProto.GetHistoryOfVersionRequest): Status {
```
dbarashev: Why is it placed here, far apart from the calling method and right in-between two other methods which are not related to this one but related to each other? 

Those who will read your code will expect related things to be grouped together -- that's natural expectation of any programmer. When programmer sees such interleaving, his WTF/second rate grows.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -76,11 +84,65 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
...
+    override fun getHistoryOfVersion(request: CosmasProto.GetHistoryOfVersionRequest,
+                                      responseObserver: StreamObserver<CosmasProto.GetHistoryOfVersionResponse>) {
+        println("Get request for history ${request.version} file # ${request.fileId}")
+        val response = CosmasProto.GetHistoryOfVersionResponse.newBuilder()
+        synchronized(this.versions) {
+            val fileHistoryOfVersions = this.versions[request.fileId]
+            val requestStatus = verifyGetHistoryOfVersionRequest(request)
+            if (requestStatus.isOk) {
+                val history = fileHistoryOfVersions?.get(request.version)
+                response.user = history?.user
```
dbarashev: If you received ok, you know for sure that history is not null. If it suddenly becomes null, it is an error which must be reported. Report it, do not silently return success with empty structure to the client.

src/main/proto/cosmas.proto
SashaOrlova: But how I can test add request if I haven't any get requests?

src/main/proto/cosmas.proto
SashaOrlova: patches add in some order, version - number in this order

dbarashev: Thanks. I don't think that thsi field is needed, because of the usage pattern, but anyway, if it is required for whatever reason, then let's name it appropriately. E.g. `patchNum` is okay. The matter is that `version` is already used in other request (and for better purpose) and if one sees two fields named "version" in different sorts of requests, it immediately raises the question "is it the same thing?". And the answer "no" is surprising.

src/main/proto/cosmas.proto
dbarashev: You don't have to use public grpc interface in the test. If you test the implementation, you can add methods which are not exposed in any interface but which can help you with accessing object state.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -19,6 +19,7 @@ import org.junit.Assert.*
 import org.junit.Test
 import io.grpc.internal.testing.StreamRecorder
 import org.junit.Before
+import javax.sound.midi.Patch
```
dbarashev: Nice :)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -83,4 +87,23 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
             this.files[request.fileId] = fileVersions
         }
     }
+
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                                        responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
```
dbarashev: Please align this argument. Either left-align with the previous one:
```
override fun createPatch(request: CosmasProto.CreatePatchRequest,
                         responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
```

or 4 spaces indent relative to the first column of the previous line 
```
override fun createPatch(request: CosmasProto.CreatePatchRequest,
    responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
```

src/main/proto/cosmas.proto
```diff
@@ -37,9 +37,21 @@ message CreateVersionRequest {
 message CreateVersionResponse {
 }
 
+message CreatePatchRequest {
+    string user = 1;
```
dbarashev: I suggest naming it `userId`. Because it is user id indeed and because the next field is called fileId (which tells that we are using "id" suffix to indicate id)

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -18,8 +18,8 @@ import com.google.protobuf.ByteString
 import org.junit.Assert.*
 import org.junit.Test
 import io.grpc.internal.testing.StreamRecorder
-import io.grpc.stub.StreamObserver
 import org.junit.Before
+import javax.sound.midi.Patch
```
dbarashev: Nice :)

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -142,4 +142,72 @@ class CosmasInMemoryServiceTest {
                 .build()
         this.service.createVersion(newVersionRequest, createVersionRecorder)
     }
+
+    private fun addPatchToService(text: String, user: String = "No name", fileId: String = "0",
+                                            time: Long = 0) {
```
dbarashev: Are you using proportional font by any chance? That could explain such strange indents. If you do, please change it to monospaced.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -83,4 +87,23 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
             this.files[request.fileId] = fileVersions
         }
     }
+
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                                        responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
+        println("Get request for create new history of file # ${request.fileId}")
```
dbarashev: This message is quite misleading.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -83,4 +87,23 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
             this.files[request.fileId] = fileVersions
         }
     }
+
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                                        responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
+        println("Get request for create new history of file # ${request.fileId}")
+        synchronized(this.patches) {
+            val historyOfFile = patches[request.fileId] ?: mutableListOf()
```
dbarashev: you really want to use the terminology consistently. If you access patches, you most likely get a patch or patch list.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -142,4 +142,72 @@ class CosmasInMemoryServiceTest {
...
+                .newBuilder()
+                .setUser(user)
+                .setFileId(fileId)
+                .setText(text)
+                .setTimeStamp(time)
+                .build()
+        this.service.createPatch(newVersionRequest, createVersionRecorder)
+    }
+
+    private fun checkCorrect(user : String, text : String, time : Long,
```
dbarashev: Please use spaces around colons consistently. Here you add a space fro both sides while just a few lines above you ass a space from the right side only. [Kotlin code style](https://kotlinlang.org/docs/reference/coding-conventions.html#horizontal-whitespace) defines reasonable rules about spaces around colons.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -142,4 +142,72 @@ class CosmasInMemoryServiceTest {
...
+                             ans : CosmasInMemoryService.Patch?) : Boolean {
+        return ans != null && ans.user == user && ans.text == text && ans.timeStamp == time
+    }
+
+    @Test
+    fun addOnePatch() {
+        addPatchToService("Hey Jude, don't make it bad.", "The Beatles", "1", 1968)
+        val ans = service.getPatch(0, "1")
+        assertEquals("The Beatles", ans?.user)
```
dbarashev: Any sense to use `checkCorrect` function here?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -142,4 +141,68 @@ class CosmasInMemoryServiceTest {
                 .build()
         this.service.createVersion(newVersionRequest, createVersionRecorder)
     }
+
+    private fun addPatchToService(text: String, user: String = "No name", fileId: String = "0", time: Long = 0) {
```
dbarashev: You always supply the value of all arguments in your tests. You can just remove the default values.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
```diff
@@ -142,4 +141,68 @@ class CosmasInMemoryServiceTest {
                 .build()
         this.service.createVersion(newVersionRequest, createVersionRecorder)
     }
+
+    private fun addPatchToService(text: String, user: String = "No name", fileId: String = "0", time: Long = 0) {
+        val createVersionRecorder: StreamRecorder<CosmasProto.CreatePatchResponse> = StreamRecorder.create()
```
dbarashev: You *really* want to use names consistently.

dbarashev (commented):
This looks good overall, but there are merge conflicts which need to be resolved. I hope you know how to deal with conflicts. Ping me if you have no idea.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
SashaOrlova: remove

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryServiceTest.kt
SashaOrlova: change

dbarashev (approved):
LGTM. You can merge it now.

Issue #18, opened 2018-04-09
Buffered "CreateVersion" for a set of files in a project


dbarashev:
We want to implement the following scenario to support real-time collaboration of many users in the same project:

1. Client (read "frontend server") sends file contents to Cosmas without creating a new file version at the moment of sending.
2. Cosmas saves or updates the actual contents of the file in memory
3. At some moment -- either when user hits "Compile" or just once per few minutes -- Cosmas commits all in-memory updates as new versions of corresponding files. 

This way we 1) will not create new versions very frequently  and 2) will save a set of possible related changes in different files in the same moment

This requires changing the behavior (and most likely the name) of CreateVersion request -- it should buffer the updates instead of writing them immediately and addign a new method CommitVersion which will actually write all buffered updates in the given project.

PR #19, opened 2018-04-09
Buffered versioning

iisuslik43:
#18 
I ll add more specific tests tomorrow.

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -30,19 +30,44 @@ import io.grpc.StatusException
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
 
+    private val fileBuffer = mutableMapOf<String, MutableMap<String, ByteString>>()
+
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         println("Get request for create new version of file # ${request.fileId}")
+
+        val project = this.fileBuffer[request.projectId] ?: mutableMapOf()
+        project[request.fileId] = request.file
+        this.fileBuffer[request.projectId] = project
+
+        val response = CosmasProto.CreateVersionResponse
+                .newBuilder()
+                .build()
+        responseObserver.onNext(response)
+        responseObserver.onCompleted()
+    }
+
+    override fun commitVersion(request: CosmasProto.CommitVersionRequest, responseObserver: StreamObserver<CosmasProto.CommitVersionResponse>) {
+        println("Get request for commit last version of files in project # ${request.projectId}")
+        val project = this.fileBuffer[request.projectId]
+        if (project == null) {
+            val status = Status.INVALID_ARGUMENT.withDescription(
+                    "There is no project in buffer with project id ${request.projectId}")
+            println(status.description)
+            responseObserver.onError(StatusException(status))
+            return
+        }
         try {
-            val blob = this.storage.create(
-                    BlobInfo.newBuilder(this.bucketName, request.fileId).build(),
-                    request.file.toByteArray())
-            println("Generation of created file: ${blob.generation}")
+            project.forEach { (fileId, file) ->
+                this.storage.create(
+                        BlobInfo.newBuilder(this.bucketName, fileId).build(),
+                        file.toByteArray())
+            }
         } catch (e: StorageException) {
             handleStorageException(e, responseObserver)
             return
         }
-        val response: CosmasProto.CreateVersionResponse = CosmasProto.CreateVersionResponse
+        val response = CosmasProto.CommitVersionResponse
                 .newBuilder()
                 .build()
         responseObserver.onNext(response)
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -35,11 +35,11 @@ class CosmasGoogleCloudService(private val bucketName: String,
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         println("Get request for create new version of file # ${request.fileId}")
-
-        val project = this.fileBuffer[request.projectId] ?: mutableMapOf()
-        project[request.fileId] = request.file
-        this.fileBuffer[request.projectId] = project
-
+        synchronized(fileBuffer) {
+            val project = this.fileBuffer[request.projectId] ?: mutableMapOf()
+            project[request.fileId] = request.file
+            this.fileBuffer[request.projectId] = project
+        }
         val response = CosmasProto.CreateVersionResponse
                 .newBuilder()
                 .build()
@@ -49,7 +49,9 @@ class CosmasGoogleCloudService(private val bucketName: String,
 
     override fun commitVersion(request: CosmasProto.CommitVersionRequest, responseObserver: StreamObserver<CosmasProto.CommitVersionResponse>) {
         println("Get request for commit last version of files in project # ${request.projectId}")
-        val project = this.fileBuffer[request.projectId]
+        val project = synchronized(fileBuffer) {
+            this.fileBuffer[request.projectId]
+        }
         if (project == null) {
             val status = Status.INVALID_ARGUMENT.withDescription(
                     "There is no project in buffer with project id ${request.projectId}")
@@ -57,15 +59,17 @@ class CosmasGoogleCloudService(private val bucketName: String,
             responseObserver.onError(StatusException(status))
             return
         }
-        try {
-            project.forEach { (fileId, file) ->
-                this.storage.create(
-                        BlobInfo.newBuilder(this.bucketName, fileId).build(),
-                        file.toByteArray())
+        synchronized(project) {
+            try {
+                project.forEach { (fileId, file) ->
+                    this.storage.create(
+                            BlobInfo.newBuilder(this.bucketName, fileId).build(),
+                            file.toByteArray())
+                }
+            } catch (e: StorageException) {
+                handleStorageException(e, responseObserver)
+                return
             }
-        } catch (e: StorageException) {
-            handleStorageException(e, responseObserver)
-            return
         }
         val response = CosmasProto.CommitVersionResponse
                 .newBuilder()
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -30,13 +30,13 @@ import io.grpc.StatusException
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
 
-    private val fileBuffer = mutableMapOf<String, MutableMap<String, ByteString>>()
+    private val fileBuffer = mutableMapOf<String, MutableMap<String, ByteString>>().withDefault { mutableMapOf() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         println("Get request for create new version of file # ${request.fileId}")
         synchronized(fileBuffer) {
-            val project = this.fileBuffer[request.projectId] ?: mutableMapOf()
+            val project = this.fileBuffer.getValue(request.projectId)
             project[request.fileId] = request.file
             this.fileBuffer[request.projectId] = project
         }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -20,6 +20,7 @@ import io.grpc.stub.StreamObserver
 import com.google.protobuf.ByteString
 import io.grpc.Status
 import io.grpc.StatusException
+import java.util.concurrent.ConcurrentHashMap
 
 /**
  * Special class that can work with requests from CosmasClient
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -30,19 +30,44 @@ import io.grpc.StatusException
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
 
+    private val fileBuffer = mutableMapOf<String, MutableMap<String, ByteString>>()
+
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         println("Get request for create new version of file # ${request.fileId}")
+
+        val project = this.fileBuffer[request.projectId] ?: mutableMapOf()
```
dbarashev: Protect fileBuffer modifications with synchronized block.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -30,19 +30,44 @@ import io.grpc.StatusException
...
+        project[request.fileId] = request.file
+        this.fileBuffer[request.projectId] = project
```
dbarashev: Any sense to use Kotlin's [`withDefault`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/with-default.html) ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -30,19 +30,44 @@ import io.grpc.StatusException
...
+        val response = CosmasProto.CreateVersionResponse
+                .newBuilder()
+                .build()
+        responseObserver.onNext(response)
+        responseObserver.onCompleted()
+    }
+
+    override fun commitVersion(request: CosmasProto.CommitVersionRequest, responseObserver: StreamObserver<CosmasProto.CommitVersionResponse>) {
+        println("Get request for commit last version of files in project # ${request.projectId}")
+        val project = this.fileBuffer[request.projectId]
```
dbarashev, line 53: Synchronize access to fileBuffer and project

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -132,22 +134,23 @@ class CosmasGoogleCloudServiceTest {
...
-                .setFile(ByteString.copyFromUtf8("text"))
-                .build()
-        this.service.createVersion(newVersionRequest, createVersionRecorder)
-        assertNotNull(createVersionRecorder.error)
-        assertEquals("test", createVersionRecorder.error!!.message)
+        addFileToService("file")
+        val commitRecorder = getStreamRecorderForCommitVersions()
+        assertNotNull(commitRecorder.error)
+        assertEquals("test", commitRecorder.error!!.message)
         this.service = getServiceForTests()
```
dbarashev: Why don't you move this "post-initialization" to `@Before` method?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -201,4 +204,19 @@ class CosmasGoogleCloudServiceTest {
         Mockito.`when`(blob.generation).thenReturn(generation)
         return blob
     }
+
+    private fun getStreamRecorderForCommitVersions(projectId: String = "0"):
```
dbarashev: It is kinda unexpected that method named `get` makes such side-effects. It is common when `get` methods are not readonly and modify state (e.g. if property is lazy-initialized) but usually `get` method won't commit anything in some other service.

I recommend leaving in this method only what its name claims -- create a stream recorder -- and move commit call to the calling method. Even if it makes the code longer, it will make the code of particular test easier to grasp, because you will see what that test does with `this.service` by looking at the test function only.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: ok, done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: ok, now this methods are returning pair of stream recorder and request and aren't doing anything alse

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I dont quite understand how to use it there.
If i add `private val fileBuffer = mutableMapOf<...>().withDefault{mutableMapOf()}`, fileBuffer[request.projectId] still nullable, so it doesn't help

dbarashev: Apparently it works for `getValue` method only:
```
    val map: Map<String, String> = mutableMapOf<String, String>().withDefault({key -> key + "!!!"})
    println(map.getValue("foo"))
```

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -30,19 +30,48 @@ import io.grpc.StatusException
...
-            val blob = this.storage.create(
-                    BlobInfo.newBuilder(this.bucketName, request.fileId).build(),
-                    request.file.toByteArray())
-            println("Generation of created file: ${blob.generation}")
-        } catch (e: StorageException) {
-            handleStorageException(e, responseObserver)
+        synchronized(fileBuffer) {
+            val project = this.fileBuffer.getValue(request.projectId)
+            project[request.fileId] = request.file
+            this.fileBuffer[request.projectId] = project
```
dbarashev: You don't need to set the value if you're using implicit defaults.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -30,19 +30,48 @@ import io.grpc.StatusException
...
+        }
+        if (project == null) {
+            val status = Status.INVALID_ARGUMENT.withDescription(
+                    "There is no project in buffer with project id ${request.projectId}")
+            println(status.description)
+            responseObserver.onError(StatusException(status))
             return
         }
-        val response: CosmasProto.CreateVersionResponse = CosmasProto.CreateVersionResponse
+        synchronized(project) {
```
dbarashev: Strictly speaking this will not work because at line 40 you modify the map outside of the critical section. You need to either move this access into previous syncronized(fileBuffer) or surround line 40 with synchronized(project) or use concurrenthashmap which is arguably the best choice :)

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -40,44 +41,48 @@ class CosmasGoogleCloudServiceTest {
 
     @Before
     fun testInitialization() {
+        service = getServiceForTests()
```
dbarashev: `this.` ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I didn't find how to use it without setting `this.fileBuffer[request.projectId] = project`

dbarashev: Oh, I see, `withDefault` just returns a value and won't  put it into a map. Not so much useful, but okay.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -29,17 +30,15 @@ import io.grpc.StatusException
...
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
-
-    private val fileBuffer = mutableMapOf<String, MutableMap<String, ByteString>>().withDefault { mutableMapOf() }
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentHashMap<String, ByteString>>().withDefault { ConcurrentHashMap() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         println("Get request for create new version of file # ${request.fileId}")
-        synchronized(fileBuffer) {
```
dbarashev: Hey, I didn't call on removal of synchronized blocks. ConcurrentHashMap is not a replacement of mutex synchronization in general, although it might be a replacement in some cases. It guarantees predictable behavior when two conflicting methods, such as entries() and put() are executed concurrently, but here you have a sequence of calls: 1)get value + 2)do something + 3)put value and ConcurrentHashMap will not help you if some concurrent thread executing the same sequence. The winner will be the one who executes 3) the last which is not what you want.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -29,20 +30,45 @@ import io.grpc.StatusException
  */
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentHashMap<String, ByteString>>().withDefault { ConcurrentHashMap() }
```
dbarashev: Any sense to replace ConcurrentHashMap in generic type with Map ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -29,20 +30,45 @@ import io.grpc.StatusException
...
+            println(status.description)
+            responseObserver.onError(StatusException(status))
+            return
+        }
         try {
-            val blob = this.storage.create(
-                    BlobInfo.newBuilder(this.bucketName, request.fileId).build(),
-                    request.file.toByteArray())
-            println("Generation of created file: ${blob.generation}")
+            project.forEach { (fileId, file) ->
```
dbarashev, line 63: This is the case when ConcurrentHashMap is useful. Regular HashMap throws ConcurrentModificationException if map is modified concurrently with iteration. It can be modified if someone calls createVersion concurrently with commitVersion, that's why I asked to do something here. But it doesn't mean that you don't need other kinds of synchronization.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -132,40 +138,87 @@ class CosmasGoogleCloudServiceTest {
...
+        createVersion("ver0")
+        createVersion("ver1")
+        commit()
+        assertEquals("ver1", getFileFromService(0))
+    }
+
+    @Test
+    fun makeTwoCommits() {
+        val fakeStorage: Storage = mock(Storage::class.java)
+        val blob1 = getMockedBlob("ver2", 0)
```
dbarashev, line 159: In `getMockedBlob` the first argument name is misleading. It it file content, not file as it says.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: But `commitVersion` doesn't write anything, so I don't have to add synchronized block, right?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -132,40 +138,87 @@ class CosmasGoogleCloudServiceTest {
...
+                any(BlobInfo::class.java), eq("ver2".toByteArray())))
+                .thenReturn(blob1)
+        Mockito.`when`(fakeStorage.create(
+                any(BlobInfo::class.java), eq("ver4".toByteArray())))
+                .thenReturn(blob2)
+        Mockito.`when`(fakeStorage.get(
+                any(BlobId::class.java), any(Storage.BlobGetOption::class.java)))
+                .thenReturn(blob1).thenReturn(blob2)
+        this.service = CosmasGoogleCloudService(this.BUCKET_NAME, fakeStorage)
+        createVersion("ver1")
```
dbarashev, line 171: Please verify that fakeStorage.create is never called with "ver1" and "ver3" arguments.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
dbarashev: That's right, in `commitVersion` you don't have to synchronize.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: `fileBuffer = ConcurrentHashMap<String, Map<...>>`? But with usual Map it will die with exception in iteration.
Or it can work, because in default it still will be ConcurrentHashMap? It's strange, because later I can make a mistake and wirte fileBuffer[...] = mutableMapOf()

dbarashev: Map is an interface, it is implemented by both HashMap and ConcurrentHashMap. Generic type doesn't care about the concrete implementation.

However, it is a good point about additional protection from error by requiring concurrent data structure. You can replace it with [ConcurrentMap](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html) interface then

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: done

dbarashev (approved):
LGTM. You can merge it now.

Issue #20, opened 2018-04-09
Replace println with logging


dbarashev:
We already have a lot of `println` statements in the code. Let's replace them with [slf4j+logback](https://www.slf4j.org/manual.html) based logging.

PR #21, opened 2018-04-11
Replace println with logging

iisuslik43:
#20

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,8 +45,8 @@ dependencies {
     compile 'com.google.cloud:google-cloud-storage:1.23.+'
     compile 'com.google.cloud:google-cloud-datastore:1.23.+'
     compile 'com.google.cloud:google-cloud-nio:0.42.+'
-    compile group: 'ch.qos.logback', name: 'logback-classic', version: '0.9.+'
-    compile group: 'ch.qos.logback', name: 'logback-core', version: '0.9.+'
+    compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.3.+'
+    compile group: 'ch.qos.logback', name: 'logback-core', version: '1.3.+'
     testCompile group: 'junit', name: 'junit', version: '4.12'
     testCompile group: 'org.mockito', name: 'mockito-all', version: '1.8.+'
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -45,8 +45,11 @@ dependencies {
     compile 'com.google.cloud:google-cloud-storage:1.23.+'
     compile 'com.google.cloud:google-cloud-datastore:1.23.+'
     compile 'com.google.cloud:google-cloud-nio:0.42.+'
+    compile group: 'ch.qos.logback', name: 'logback-classic', version: '0.9.+'
```
dbarashev: Apparently 0.9.+ is too rusty: https://mvnrepository.com/artifact/ch.qos.logback/logback-classic

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasClient.kt
```diff
@@ -21,30 +21,31 @@ import com.xenomachina.argparser.ArgParser
...
+import org.slf4j.LoggerFactory
 import java.util.concurrent.TimeUnit
 
 /**
  * Simple client that will send a request to Cosmas server and wait for response
  * @author Aleksandr Fedotov (iisuslik43)
  */
 class CosmasClient(host: String, port: Int) {
-
+    private val log = LoggerFactory.getLogger(this::class.java)
```
dbarashev: This looks a bit dangerous because moving this code will change the effective logger name as well. Normally one would use explicit class reference (CosmasClient::class.java) or even strings (e.g. "CosmasClient") to avoid long logger names.

Also, please capitalize `log` and even move it outside of the class, so that it could be used without `this.` prefix and it was clear it is a logger, not some local var:
`LOG.info`

iisuslik43: But if I move `log` outside and do it with all files in package, it will be conflicting declarations, so maybe I should make it static.

dbarashev: If you make them private val you will have no conflicts.
I just've googled kotlin logging and oh dear how many interesting solutions are out there.

build.gradle
iisuslik43: done

dbarashev (approved):
LGTM

Issue #22, opened 2018-04-11
Add patch list to the stored file content versions


dbarashev:
Here is a summary of process which we can support now:
```
Time          t1      t2      t3        t4     t5      t6 
Alice types   "foo"                            "baz"
Bob   types           "bar"             "qux"      
Cosmas does   +patch  +patch  +version  +patch +patch  +version
              "foo"   "bar"   "foobar"  "qux"  "baz"   "foobar quxbaz"
```

We have versions and we have a separate patch list. However, we need them to be synchronized, because we want to support the following user scenario: Alice takes version produced at time t3 and applies patches one-by-one.

Arguably the best way to get consistent version and patch list to the next version is to keep patch list and version together in the same structure. 

So, the plan: replace file contents which is written to GCS on commit version request with file contents + accumulated patch list structure (which might be a protocol buffer serialized to bytes). Then we empty the patch lists for the committed files and start accumulating patches again.

PR #23, opened 2018-04-15
patchs to file structure

SashaOrlova:
#22 
I don't understand how to test it better

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -16,12 +16,10 @@ package com.bardsoftware.papeeria.backend.cosmas
 
 import com.google.api.gax.paging.Page
 import com.google.cloud.storage.*
-import com.google.protobuf.ByteString
 import io.grpc.Status
 import io.grpc.StatusException
 import io.grpc.stub.StreamObserver
 import org.slf4j.LoggerFactory
-import java.io.Serializable
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 
@@ -36,14 +34,14 @@ private val LOG = LoggerFactory.getLogger("CosmasGoogleCloudService")
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
 
-    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileInformation>>().withDefault { ConcurrentHashMap() }
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileVersion>>().withDefault { ConcurrentHashMap() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         LOG.info("Get request for create new version of file # ${request.fileId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            project[request.fileId] = CosmasProto.FileInformation.newBuilder().setFile(request.file).addAllPatches(mutableListOf()).build()
+            project[request.fileId] = CosmasProto.FileVersion.newBuilder().setContent(request.file).addAllPatches(project[request.fileId]?.patchesList ?: mutableListOf()).build()
             this.fileBuffer[request.projectId] = project
         }
         val response = CosmasProto.CreateVersionResponse
@@ -55,15 +53,12 @@ class CosmasGoogleCloudService(private val bucketName: String,
 
     override fun createPatch(request: CosmasProto.CreatePatchRequest,
                              responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
-        LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.userId}")
+        LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.patch.userId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            val newPatch = CosmasProto.FileInformation.Patch.newBuilder().setText(request.text)
-                    .setTimeStamp(request.timeStamp).setUserId(request.userId).build()
             val fileInformation = project[request.fileId]
             if (fileInformation != null) {
-                project[request.fileId] = CosmasProto.FileInformation.newBuilder()
-                        .setFile(fileInformation.file).addAllPatches(fileInformation.patchesList).addPatches(newPatch).build()
+                project[request.fileId] = fileInformation.toBuilder().addPatches(request.patch).build()
                 this.fileBuffer[request.projectId] = project
             }
         }
@@ -85,10 +80,11 @@ class CosmasGoogleCloudService(private val bucketName: String,
             return
         }
         try {
-            for ((fileId, fileStorage) in project) {
+            for ((fileId, fileVersion) in project) {
                 this.storage.create(
                         BlobInfo.newBuilder(this.bucketName, fileId).build(),
-                        fileStorage.toByteArray())
+                        fileVersion.toByteArray())
+                project[fileId] = fileVersion.toBuilder().clearPatches().build()
             }
         } catch (e: StorageException) {
             handleStorageException(e, responseObserver)
@@ -120,7 +116,7 @@ class CosmasGoogleCloudService(private val bucketName: String,
             responseObserver.onError(StatusException(requestStatus))
             return
         }
-        response.file = CosmasProto.FileInformation.parseFrom(blob.getContent()).file
+        response.file = CosmasProto.FileVersion.parseFrom(blob.getContent()).content
         responseObserver.onNext(response.build())
         responseObserver.onCompleted()
     }
@@ -164,17 +160,17 @@ class CosmasGoogleCloudService(private val bucketName: String,
         }
     }
 
-    fun getPatchList(projectId: String, fileId: String): MutableList<CosmasProto.FileInformation.Patch>? {
+    fun getPatchList(projectId: String, fileId: String): List<CosmasProto.Patch>? {
         return fileBuffer[projectId]?.get(fileId)?.patchesList
     }
 
-    fun getPatchListFromMemory(fileId: String, version: Long): MutableList<CosmasProto.FileInformation.Patch>? {
+    fun getPatchListFromStorage(fileId: String, version: Long): List<CosmasProto.Patch>? {
         val blob: Blob? = try {
             this.storage.get(BlobInfo.newBuilder(this.bucketName, fileId).build().blobId,
                     Storage.BlobGetOption.generationMatch(version))
         } catch (e: StorageException) {
             return null
         }
-        return CosmasProto.FileInformation.parseFrom(blob?.getContent()).patchesList
+        return CosmasProto.FileVersion.parseFrom(blob?.getContent()).patchesList
     }
 }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -21,11 +21,12 @@ import io.grpc.Status
 import io.grpc.StatusException
 import io.grpc.stub.StreamObserver
 import org.slf4j.LoggerFactory
+import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
+import java.io.ObjectInputStream
 import java.io.ObjectOutputStream
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
-
 
 private val LOG = LoggerFactory.getLogger("CosmasGoogleCloudService")
 
@@ -38,15 +39,15 @@ private val LOG = LoggerFactory.getLogger("CosmasGoogleCloudService")
 class CosmasGoogleCloudService(private val bucketName: String,
                                private val storage: Storage = StorageOptions.getDefaultInstance().service) : CosmasGrpc.CosmasImplBase() {
 
-    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, ByteString>>().withDefault { ConcurrentHashMap() }
-    private val patchBuffer = ConcurrentHashMap<String, ConcurrentMap<String, MutableList<CosmasInMemoryService.Patch>>>().withDefault { ConcurrentHashMap() }
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String,
+            Pair<ByteString, MutableList<CosmasInMemoryService.Patch>>>>().withDefault { ConcurrentHashMap() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         LOG.info("Get request for create new version of file # ${request.fileId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            project[request.fileId] = request.file
+            project[request.fileId] = Pair(request.file, mutableListOf())
             this.fileBuffer[request.projectId] = project
         }
         val response = CosmasProto.CreateVersionResponse
@@ -59,12 +60,13 @@ class CosmasGoogleCloudService(private val bucketName: String,
     override fun createPatch(request: CosmasProto.CreatePatchRequest,
                              responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
         LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.userId}")
-        synchronized(this.patchBuffer) {
-            val project = this.patchBuffer.getValue(request.projectId)
-            val patchesList = project[request.fileId] ?: mutableListOf()
+        synchronized(this.fileBuffer) {
+            val project = this.fileBuffer.getValue(request.projectId)
+            val fileInformation = project[request.fileId]
+            val patchesList = fileInformation?.second ?: mutableListOf()
             patchesList.add(CosmasInMemoryService.Patch(request.userId, request.text, request.timeStamp))
-            project[request.fileId] = patchesList
-            this.patchBuffer[request.projectId] = project
+            project[request.fileId] = Pair(fileInformation?.first ?: ByteString.EMPTY, patchesList)
+            this.fileBuffer[request.projectId] = project
         }
         val response: CosmasProto.CreatePatchResponse = CosmasProto.CreatePatchResponse
                 .newBuilder()
@@ -76,26 +78,6 @@ class CosmasGoogleCloudService(private val bucketName: String,
     override fun commitVersion(request: CosmasProto.CommitVersionRequest, responseObserver: StreamObserver<CosmasProto.CommitVersionResponse>) {
         LOG.info("Get request for commit last version of files in project # ${request.projectId}")
         val project = this.fileBuffer[request.projectId]
-        val patchProject = this.patchBuffer[request.projectId]
-
-        if (patchProject != null) {
-            try {
-                for ((fileId, patchList) in patchProject) {
-                    val outputStream = ByteArrayOutputStream()
-                    val output = ObjectOutputStream(outputStream)
-                    for (patch in patchList) {
-                        output.writeObject(patch)
-                    }
-                    this.storage.create(
-                            BlobInfo.newBuilder(this.bucketName, fileId).build(),
-                            outputStream.toByteArray())
-                    patchList.clear()
-                }
-            } catch (e: StorageException) {
-                handleStorageException(e, responseObserver)
-                return
-            }
-        }
         if (project == null) {
             val status = Status.INVALID_ARGUMENT.withDescription(
                     "There is no project in buffer with project id ${request.projectId}")
@@ -104,10 +86,14 @@ class CosmasGoogleCloudService(private val bucketName: String,
             return
         }
         try {
-            project.forEach { (fileId, file) ->
+            for ((fileId, pair) in project) {
+                val outputStream = ByteArrayOutputStream()
+                val output = ObjectOutputStream(outputStream)
+                output.writeObject(pair)
+                output.flush()
                 this.storage.create(
                         BlobInfo.newBuilder(this.bucketName, fileId).build(),
-                        file.toByteArray())
+                        outputStream.toByteArray())
             }
         } catch (e: StorageException) {
             handleStorageException(e, responseObserver)
@@ -139,7 +125,12 @@ class CosmasGoogleCloudService(private val bucketName: String,
             responseObserver.onError(StatusException(requestStatus))
             return
         }
-        response.file = ByteString.copyFrom(blob.getContent())
+        val ans = blob.getContent()
+        val inputStream = ByteArrayInputStream(ans)
+        val input = ObjectInputStream(inputStream)
+        val pair = input.readObject()
+        if (pair is Pair<*, *> && pair.first is ByteString)
+            response.file = pair.first as ByteString
         responseObserver.onNext(response.build())
         responseObserver.onCompleted()
     }
@@ -184,6 +175,22 @@ class CosmasGoogleCloudService(private val bucketName: String,
     }
 
     fun getPatchList(projectId: String, fileId: String): MutableList<CosmasInMemoryService.Patch>? {
-        return patchBuffer[projectId]?.get(fileId)
+        return fileBuffer[projectId]?.get(fileId)?.second
+    }
+
+    fun getPatchListFromMemory(fileId: String, version: Long): MutableList<CosmasInMemoryService.Patch>? {
+        val blob: Blob? = try {
+            this.storage.get(BlobInfo.newBuilder(this.bucketName, fileId).build().blobId,
+                    Storage.BlobGetOption.generationMatch(version))
+        } catch (e: StorageException) {
+            return null
+        }
+        val ans = blob?.getContent()
+        val inputStream = ByteArrayInputStream(ans)
+        val input = ObjectInputStream(inputStream)
+        val pair = input.readObject()
+        if (pair is Pair<*, *> && pair.second is MutableList<*>)
+            return pair.second as MutableList<CosmasInMemoryService.Patch>?
+        return null
     }
 }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -25,6 +25,7 @@ import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
 import java.io.ObjectInputStream
 import java.io.ObjectOutputStream
+import java.io.Serializable
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -52,10 +56,46 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
 
+        if (patchProject != null) {
+            try {
+                for ((fileId, patchList) in patchProject) {
+                    val outputStream = ByteArrayOutputStream()
+                    val output = ObjectOutputStream(outputStream)
+                    for (patch in patchList) {
+                        output.writeObject(patch)
+                    }
+                    this.storage.create(
```
dbarashev: Come on. You write messages with different meaning (patch list vs file version) to the same object. This is a wrong way.

dbarashev (commented):
Hey,
this could work after fixing pretty obvious flaws, but we actually want more deep integration between file contents and patch list. 

With the schema proposed in this pull request if one wants to extract a list of patches which will make a transition from version1 to version2 (e.g. from time t3 to time t6 in the example in the issue), he will have to 
1. Get version 2 from GCS
2. Find out the range of timestamps or ids of pacthes
3. Fetch patches one-by-one

Totally, it could require non-trivial logic and O(number of patches) accesses to GCS which is expensive from both time and money points.

What we really want is to be able to fetch file contents of version N and list of patches from version N-1 to N in a single call to GCS. For that purpose the list of patches should sit in the same data structure with fiel contents, that is, in a protocol buffer.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -35,14 +36,14 @@ private val LOG = LoggerFactory.getLogger("CosmasGoogleCloudService")
...
-    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, ByteString>>().withDefault { ConcurrentHashMap() }
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileInformation>>().withDefault { ConcurrentHashMap() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         LOG.info("Get request for create new version of file # ${request.fileId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            project[request.fileId] = request.file
+            project[request.fileId] = CosmasProto.FileInformation.newBuilder().setFile(request.file).addAllPatches(mutableListOf()).build()
```
dbarashev: You're going to reset the patch list every time when `createVersion` is called. It is wrong, because we may call `createVersion` many times in a row between two consecutive `commitVersion` calls. You want to keep the existing patch list until committing.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -52,10 +53,30 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                             responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
+        LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.userId}")
+        synchronized(this.fileBuffer) {
+            val project = this.fileBuffer.getValue(request.projectId)
+            val newPatch = CosmasProto.FileInformation.Patch.newBuilder().setText(request.text)
+                    .setTimeStamp(request.timeStamp).setUserId(request.userId).build()
+            val fileInformation = project[request.fileId]
+            if (fileInformation != null) {
+                project[request.fileId] = CosmasProto.FileInformation.newBuilder()
```
dbarashev: I think that `fileInformation.toBuilder()` is more canonical and efficient way of making a mutable copy.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -64,10 +85,10 @@ class CosmasGoogleCloudService(private val bucketName: String,
             return
         }
         try {
-            project.forEach { (fileId, file) ->
+            for ((fileId, fileStorage) in project) {
```
dbarashev: You know, you call it `fileStorage` here and `fileInformation` in `createPatch` method. Please choose a single name, otherwise we're all going to suffer from light schizophrenia. 
Usually if class is called TheThing then you call variables `thatThing`, `thisThing`, `myThing` etc. However, as I wrote in one of my essays, I personally don't like the name `FileInformation` because it basically means nothing. Perhaps `FileVersion` with `contents` and `patches` fields is more appropriate.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -142,4 +163,18 @@ class CosmasGoogleCloudService(private val bucketName: String,
             LOG.info("Deleting file failed: ${e.message}")
         }
     }
+
+    fun getPatchList(projectId: String, fileId: String): MutableList<CosmasProto.FileInformation.Patch>? {
+        return fileBuffer[projectId]?.get(fileId)?.patchesList
+    }
+
+    fun getPatchListFromMemory(fileId: String, version: Long): MutableList<CosmasProto.FileInformation.Patch>? {
```
dbarashev: Alexandra, normally when programmer sees the words "from memory" he assumes that it is RAM. It is pretty surprising that a method which claims to get something from memory actually accesses persistent storage.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -142,4 +163,18 @@ class CosmasGoogleCloudService(private val bucketName: String,
             LOG.info("Deleting file failed: ${e.message}")
         }
     }
+
+    fun getPatchList(projectId: String, fileId: String): MutableList<CosmasProto.FileInformation.Patch>? {
```
dbarashev: Normally you'll want to return immutable collections, unless you really want to tell the caller that it is okay to modify the collection directly (it is not the case)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
```diff
@@ -33,7 +34,7 @@ class CosmasInMemoryService : CosmasGrpc.CosmasImplBase() {
     private val files = mutableMapOf<String, MutableList<ByteString>>()
     private val patches = mutableMapOf<String, MutableList<Patch>>()
 
-    data class Patch(val user: String, val text: String, val timeStamp: Long)
+    data class Patch(val user: String, val text: String, val timeStamp: Long) : Serializable
```
dbarashev: serializable not needed anymore?

src/main/proto/cosmas.proto
```diff
@@ -64,6 +65,16 @@ message CommitVersionResponse {
 
 }
 
+message FileInformation {
+    message Patch {
```
dbarashev: any sense to extract this Patch message to top-level and use it both here and in `CreatePatchRequest`?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +201,59 @@ class CosmasGoogleCloudServiceTest {
         assertNotNull(stream.error)
     }
 
+    @Test
+    fun simpleAddPatch() {
+        createVersion("file", "1", "1")
+        addPatchToService("abc", "-", "1", 1, "1")
+        commit("1")
```
dbarashev: don't you want to check something in the test? How do you know that implementations of these methods are not empty?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +201,59 @@ class CosmasGoogleCloudServiceTest {
...
+    }
+
+    @Test
+    fun checkEmptyList() {
+        createVersion("file", "1", "1")
+        addPatchToService("abc", "-", "1", 1, "1")
+        commit("1")
+        createVersion("file1", "1", "1")
+        val list = this.service.getPatchList("1", "1")
+        assertEquals(0, list?.size)
```
dbarashev: you are not sure if you get null or not? Test is essentially a specification, it says how code is expected to work. If you expect that `list` is null after this call, assert that it is null. If you expect that it is not null then assert that it is not null (you can use Kotlin's `!!` operator for that).

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +201,59 @@ class CosmasGoogleCloudServiceTest {
...
+        createVersion("file", "1", "1")
+        addPatchToService("abc", "-", "1", 1, "1")
+        addPatchToService("abc1", "-1", "1", 12, "1")
+        commit("1")
+        createVersion("file2", "1", "1")
+        addPatchToService("abcd", "-d", "1", 13, "1")
+        addPatchToService("abcd1", "-d1", "1", 14, "1")
+        commit("1")
+        assertEquals(getFileFromService(0, "1", "1"), getFileFromService(1, "1", "1"))
+        val list = this.service.getPatchListFromMemory("1", 1)
```
dbarashev: any sense to check that patch list from the previous version is also available?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +201,59 @@ class CosmasGoogleCloudServiceTest {
...
+    fun getPatchManyVersions() {
+        createVersion("file", "1", "1")
+        addPatchToService("abc", "-", "1", 1, "1")
+        addPatchToService("abc1", "-1", "1", 12, "1")
+        commit("1")
+        createVersion("file2", "1", "1")
+        addPatchToService("abcd", "-d", "1", 13, "1")
+        addPatchToService("abcd1", "-d1", "1", 14, "1")
+        commit("1")
+        assertEquals(getFileFromService(0, "1", "1"), getFileFromService(1, "1", "1"))
```
dbarashev: In my opinion it makes more sense to actually change file contents, just to show the use case which is closer to the reality. In the real life you don't expect file contents to stay the same after adding some patches, do you? Well, unless the patches annihilate each other. So should be in the test. On the contrary, if you waste your time and typing on assertion that contents remains the same then  you probably want to say something to the reader, and the reader starts suffering from the cognitive dissonance.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok, change to FileVersion

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok, change name to ...FromStorage

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasInMemoryService.kt
SashaOrlova: ok

src/main/proto/cosmas.proto
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: change tests

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -35,14 +34,14 @@ private val LOG = LoggerFactory.getLogger("CosmasGoogleCloudService")
...
-    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, ByteString>>().withDefault { ConcurrentHashMap() }
+    private val fileBuffer = ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileVersion>>().withDefault { ConcurrentHashMap() }
 
     override fun createVersion(request: CosmasProto.CreateVersionRequest,
                                responseObserver: StreamObserver<CosmasProto.CreateVersionResponse>) {
         LOG.info("Get request for create new version of file # ${request.fileId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            project[request.fileId] = request.file
+            project[request.fileId] = CosmasProto.FileVersion.newBuilder().setContent(request.file).addAllPatches(project[request.fileId]?.patchesList ?: mutableListOf()).build()
```
dbarashev: Please extract `project[request.fileId]` into local val and replace this chain of safe calls with simple `if`. Chains are nice until they become a mess.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -52,10 +51,27 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                             responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
+        LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.patch.userId}")
+        synchronized(this.fileBuffer) {
+            val project = this.fileBuffer.getValue(request.projectId)
+            val fileInformation = project[request.fileId]
+            if (fileInformation != null) {
```
dbarashev: What if it is null? If you believe that it is illegal state, please send the error (you didn't add a patch as client requested => you need to inform the client) or at least log it as error. 

However, I can imagine that such state is actually legal, e.g. when two GRPC requests, one `createPatch` and another `createVersion` are issued in parallel and the former comes to Cosmas earlier than the latter.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -52,10 +51,27 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    override fun createPatch(request: CosmasProto.CreatePatchRequest,
+                             responseObserver: StreamObserver<CosmasProto.CreatePatchResponse>) {
+        LOG.info("Get request for create new patch of file # ${request.fileId} by user ${request.patch.userId}")
+        synchronized(this.fileBuffer) {
+            val project = this.fileBuffer.getValue(request.projectId)
+            val fileInformation = project[request.fileId]
```
dbarashev: We agreed that it should be `fileVersion`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -64,10 +80,11 @@ class CosmasGoogleCloudService(private val bucketName: String,
             return
         }
         try {
-            project.forEach { (fileId, file) ->
+            for ((fileId, fileVersion) in project) {
                 this.storage.create(
                         BlobInfo.newBuilder(this.bucketName, fileId).build(),
-                        file.toByteArray())
+                        fileVersion.toByteArray())
+                project[fileId] = fileVersion.toBuilder().clearPatches().build()
```
dbarashev: Thanks for pointing this out. I think it is okay to just clear the map when the loop finishes.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +200,118 @@ class CosmasGoogleCloudServiceTest {
         assertNotNull(stream.error)
     }
 
+    @Test
+    fun simpleAddPatch() {
+        createVersion("file", "1", "1")
+        addPatchToService("abc", "-", "1", 1, "1")
+        val list = this.service.getPatchList("1", "1")
+        assertFalse(list == null)
```
dbarashev: `assertNotNull` is more canonical way

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +200,118 @@ class CosmasGoogleCloudServiceTest {
...
+        val list1 = this.service.getPatchList("1", "1")
+        assertFalse(list1 == null)
+        assertEquals(1, list1!!.size)
+        assertEquals("abcNew", list1[0].text)
+        assertEquals("-New", list1[0].userId)
+        assertEquals(10L, list1[0].timeStamp)
+    }
+
+    @Test
+    fun checkEmptyList() {
```
dbarashev, line 227: Cool test, thanks. Can you add similar check that version in memory becomes null wfter committing?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -200,6 +200,118 @@ class CosmasGoogleCloudServiceTest {
...
+        val listPatch2 = mutableListOf<Patch>()
+        val patch1 = CosmasProto.Patch.newBuilder().setText("patch1").setUserId("-1").setTimeStamp(1).build()
+        val patch2 = CosmasProto.Patch.newBuilder().setText("patch2").setUserId("-2").setTimeStamp(2).build()
+        val patch3 = CosmasProto.Patch.newBuilder().setText("patch3").setUserId("-3").setTimeStamp(3).build()
+        val patch4 = CosmasProto.Patch.newBuilder().setText("patch4").setUserId("-4").setTimeStamp(4).build()
+        listPatch1.add(patch1)
+        listPatch1.add(patch2)
+        listPatch2.add(patch3)
+        listPatch2.add(patch4)
+        val fileVersion1 = CosmasProto.FileVersion.newBuilder().setContent(ByteString.copyFrom("ver1".toByteArray())).addAllPatches(listPatch1).build()
```
dbarashev: Please split such long chains with linebreaks before dots:
```
CosmasProto.FileVersion.newBuilder()
    .setContent(...)
    .addAllPatches(...)
    .build()
```

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -253,7 +365,15 @@ class CosmasGoogleCloudServiceTest {
 
     private fun getMockedBlob(fileContent: String, generation: Long = 0): Blob {
         val blob = mock(Blob::class.java)
-        Mockito.`when`(blob.getContent()).thenReturn(fileContent.toByteArray())
+        Mockito.`when`(blob.getContent()).thenReturn(FileVersion.newBuilder().setContent(ByteString.copyFrom(fileContent.toByteArray())).build().toByteArray())
```
dbarashev: Please format with linebreaks

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -265,12 +385,28 @@ class CosmasGoogleCloudServiceTest {
...
-
         return Pair(commitRecorder, commitRequest)
     }
 
     private fun commit(projectId: String = "0") {
         val (commitRecorder, commitRequest) = getStreamRecorderAndRequestForCommitVersions(projectId)
         this.service.commitVersion(commitRequest, commitRecorder)
     }
+
+    private fun addPatchToService(text: String, user: String, fileId: String, time: Long, projectId: String) {
```
dbarashev: Please keep the names consistent. If you use `id` suffix to indicate that it is identifier, use it always. Otherwise your reader may think that `user` is not actually an id (name maybe?)

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -265,12 +385,28 @@ class CosmasGoogleCloudServiceTest {
...
+        val newPatchRequest = CosmasProto.CreatePatchRequest
+                .newBuilder()
+                .setFileId(fileId)
+                .setProjectId(projectId)
+                .setPatch(newPatch)
+                .build()
+        this.service.createPatch(newPatchRequest, createPatchRecorder)
+    }
+
+    private fun checkCorrect(user: String, text: String, time: Long, ans: CosmasProto.Patch?) : Boolean {
```
dbarashev: I recommend creating a new Patch from arguments (you can do it either directly in the test case code or create a helper method like this) and using `assertEquals(newPatch(userId, text, timestamp), list1[0])`

This is more self-describing way: the reader sees that test expects equality and doesn't need to learn what is "correctness" in test understanding.

dbarashev: Please. be. consistent. userId. timeStamp. 
Good read on this and related things: https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I don't see any sense to make project[request.fileId] local val, because it use in one place. I think, that it will be better to make part of chain local val

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ок

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: fix

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: add

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
SashaOrlova: ok

SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -41,7 +41,11 @@ class CosmasGoogleCloudService(private val bucketName: String,
         LOG.info("Get request for create new version of file # ${request.fileId}")
         synchronized(this.fileBuffer) {
             val project = this.fileBuffer.getValue(request.projectId)
-            project[request.fileId] = CosmasProto.FileVersion.newBuilder().setContent(request.file).addAllPatches(project[request.fileId]?.patchesList ?: mutableListOf()).build()
+            val patchList = project[request.fileId]?.patchesList ?: mutableListOf()
+            project[request.fileId] = CosmasProto.FileVersion.newBuilder()
```
dbarashev: Yup, this looks way better, thanks.

dbarashev (approved):
LGTM. You can merge it now.

PR #24, opened 2018-05-15
add possibility delete patch

SashaOrlova:


src/main/java/diff_match_patch.java
```diff
@@ -0,0 +1,2468 @@
+/*
```
dbarashev: Why did you remove package statement?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
+package com.bardsoftware.papeeria.backend.cosmas
```
dbarashev, line 16: Please add copyright notice.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
...
+import java.util.LinkedList
+import diff_match_patch
+
+class PatchCorrector {
+    private fun getReverseOperation(operation: diff_match_patch.Operation): diff_match_patch.Operation {
+        if (operation == diff_match_patch.Operation.DELETE)
```
dbarashev: Please always surround if body with curly braces. Another option for this particular case is using Kotlin's [when operator](https://kotlinlang.org/docs/reference/control-flow.html) (you can even write `return when ...`)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
...
+            reversePatchList.add(newPatch)
+        }
+        return reversePatchList
+    }
+
+    fun deletePatch(patchForDelete: LinkedList<diff_match_patch.Patch>,
+                    patchList: List<diff_match_patch.Patch>, text: String): LinkedList<diff_match_patch.Patch> {
+        val dmp = diff_match_patch()
+        var currentVersionOfText = text
+        val listOfPatches = LinkedList<diff_match_patch.Patch>()
```
dbarashev: Naming remark: you will hardly find any native or fluent English speaking programmer who uses preposition "of" in such constructs. Almost always they would use adjunct nouns: fileVersionList. It is shorter and less cluttered with short capitalized words.

Read more about adjunct nouns vs prepositions:
https://english.stackexchange.com/questions/283410/a-of-b-or-b-a-noun-adjunct-vs-prepositional-phrase
https://english.stackexchange.com/questions/225909/of-the-vs-noun-adjunct

Please remove those "Of" everywhere where it makes sense.

Besides, sometimes it is either not interesting whether a collection is a list or something else, or it is just obvious from the context. You can thus rename "listOfPatches" to "patches" or even better give it more descriptive name which would explain the difference between "listOfPatches" and "patchList".

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
...
+        var listOfReversePatches = LinkedList<diff_match_patch.Patch>()
+        listOfReversePatches.addAll(reversePatch(patchForDelete))
+        for (patch in patchList) {
+            listOfPatches.clear()
+            listOfPatches.add(patch)
```
dbarashev: Maybe there is just no difference between "listOfPatches" and "patchList" ? Why do you need the former? The only thing which you do is adding a single element after clearing. If you just need a list to pass to path_apply then Kotlin's function [listOf](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/list-of.html) is your friend.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
+package com.bardsoftware.papeeria.backend.cosmas
+import java.util.LinkedList
+import diff_match_patch
+
+class PatchCorrector {
```
dbarashev: This class has no state, which basically means that all its methods can be made static. There are many options of doing it but perhaps the easiest is to use [singleton object declaration](https://kotlinlang.org/docs/reference/object-declarations.html)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
...
+    private fun getReverseOperation(operation: diff_match_patch.Operation): diff_match_patch.Operation {
+        if (operation == diff_match_patch.Operation.DELETE)
+            return diff_match_patch.Operation.INSERT
+        if (operation == diff_match_patch.Operation.INSERT)
+            return diff_match_patch.Operation.DELETE
+        return operation
+    }
+
+    fun reversePatch(patchList: LinkedList<diff_match_patch.Patch>): LinkedList<diff_match_patch.Patch> {
+        val reversePatchList = LinkedList<diff_match_patch.Patch>()
```
dbarashev: use Kotlin's [mutableListOf](https://kotlinlang.org/docs/reference/object-declarations.html).
I don't think you need LinkedList for these purposes.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,59 @@
...
+        var listOfReversePatches = LinkedList<diff_match_patch.Patch>()
+        listOfReversePatches.addAll(reversePatch(patchForDelete))
+        for (patch in patchList) {
+            listOfPatches.clear()
+            listOfPatches.add(patch)
+            val nextVersion = dmp.patch_apply(listOfPatches, currentVersionOfText)
+            val nextVersionWithoutPatch = dmp.patch_apply(listOfReversePatches, nextVersion[0] as String)
+            var successful = true
+            for (result in nextVersion[1] as BooleanArray) {
+                successful = successful && result
```
dbarashev: You can just throw your exception the first time you see false. You will do it anyway, as soon as there is at least one false in the bitmap. 

Another option is using Kotlin's [any](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/any.html):
```
(nextVersion[1] as BooleanArray).any { !it }
```

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
```diff
@@ -0,0 +1,109 @@
+package com.bardsoftware.papeeria.backend.cosmas
```
dbarashev, line 16: Copyright notice please

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
```diff
@@ -0,0 +1,109 @@
...
+class PatchCorrectorTest {
+    @Test
+    fun plusReversePatchTest() {
+        val text1 = "Hello world!"
+        val text2 = "Hello, world!"
+        val dmp = diff_match_patch()
+        val patch = dmp.patch_make(text1, text2)
+        val patchCorrector = PatchCorrector()
+        val listOfPatch = LinkedList<diff_match_patch.Patch>()
+        listOfPatch.addAll(patchCorrector.reversePatch(patch))
```
dbarashev: Use [listOf](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/list-of.html)

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        patch.addAll(dmp.patch_make(text4, text5))
+        val patchCorrector = PatchCorrector()
+        val deletePatch = patchCorrector.deletePatch(patchForDelete, patch, text2)
+        val res = dmp.patch_apply(deletePatch, text5)
+        assertEquals("Hello, beautiful life! Bye!", res[0])
+    }
+
+    @Test
+    fun bigFileChangeWorld() {
+        val text1 = "Alice was beginning to get very tired of sitting by her sister on the bank, and of having nothing " +
```
dbarashev: Kotlin supports [multi-line string literals](https://kotlinlang.org/docs/reference/basic-types.html). You don't have to deal with concatenation anymore.

src/main/java/diff_match_patch.java
SashaOrlova: add

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: add

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: add when operator

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: delete 'Of'

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: listOf returns MutableList, but diff-match-patch use LinkedList

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: listOf returns MutableList, but diff-match-patch use LinkedList

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: ок

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
SashaOrlova: add

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
SashaOrlova: listOf returns MutableList, but diff-match-patch use LinkedList

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
SashaOrlova: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
```diff
@@ -71,39 +82,37 @@ class PatchCorrectorTest {
...
-                " pictures or conversation?’"
-        val text4 = "Ann was very tired : once or twice she had peeped into the book her sister was reading," +
-                " but it had no pictures or conversations in it, ‘and what is the use of a book,’ thought Ann ‘without" +
-                " pictures or conversation?’"
-        val text5 = "Ann was very tired and upset: once or twice she had peeped into the book her sister was reading," +
-                " but it had no pictures or conversations in it, ‘and what is the use of a book,’ thought Ann ‘without" +
-                " pictures or conversation?’"
+        val text1 = """"Alice was beginning to get very tired of sitting by her sister on the bank, and of having nothing
+            | to do: once or twice she had peeped into the book her sister was reading, but it had no pictures or conversations
+            | in it, ‘and what is the use of a book,’ thought Alice ‘without pictures or conversation?’""".trimMargin().replace("\n","")
```
dbarashev: Out of curiosity, what's the point of using such long string? Maybe really keep a few lines? Or add another test with multiline texts?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
```diff
@@ -71,39 +82,37 @@ class PatchCorrectorTest {
...
-        val text3 =  "Ann was beginning to get very tired : once or twice she had peeped into the book her sister was reading," +
-                " but it had no pictures or conversations in it, ‘and what is the use of a book,’ thought Ann ‘without" +
-                " pictures or conversation?’"
-        val text4 = "Ann was very tired : once or twice she had peeped into the book her sister was reading," +
-                " but it had no pictures or conversations in it, ‘and what is the use of a book,’ thought Ann ‘without" +
-                " pictures or conversation?’"
-        val text5 = "Ann was very tired and upset: once or twice she had peeped into the book her sister was reading," +
-                " but it had no pictures or conversations in it, ‘and what is the use of a book,’ thought Ann ‘without" +
-                " pictures or conversation?’"
+        val text1 = """"Alice was beginning to get very tired of sitting by her sister on the bank, and of having nothing
```
dbarashev: Did you really mean to use four double-quotes here?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
SashaOrlova: add multiline tests

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrectorTest.kt
SashaOrlova: fix

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -0,0 +1,69 @@
...
+                diff_match_patch.Operation.EQUAL
+            }
+        }
+    }
+
+    fun reversePatch(patchList: LinkedList<diff_match_patch.Patch>): LinkedList<diff_match_patch.Patch> {
+        val reversePatchList = LinkedList<diff_match_patch.Patch>()
+        for (patch in patchList) {
+            val newPatch = diff_match_patch.Patch()
+            newPatch.length1 = patch.length2
```
dbarashev: Are you sure it is the right way of building the reverse patch? I am not. Here is a modified version of one of your tests where I added printing of patches:
```
    fun plusReversePatchTest() {
        val text1 = "Hello world!"
        val text2 = "Hello, world!"
        val dmp = diff_match_patch()
        val patch = dmp.patch_make(text1, text2)
      println("Original patch:\n $patch")
        val listOfPatch = LinkedList<diff_match_patch.Patch>()
        listOfPatch.addAll(PatchCorrector.reversePatch(patch))
      println("Home-grown reverse:\n $listOfPatch")
      println("Reverse built by DMP:\n ${dmp.patch_make(text2, text1)}")
        val newText = dmp.patch_apply(listOfPatch, text2)
        assertEquals(text1, newText[0])
    }
```

Here is the output:
```
Original patch:
 [@@ -1,12 +1,13 @@
 Hello
+,
  world!
]
Home-grown reverse:
 [@@ -1,13 +1,12 @@
 Hello
-,
  world!
]
Reverse built by DMP:
 [@@ -2,9 +2,8 @@
 ello
-,
  wor
]
```

You can see that home-grown reverse patch is significantly different from what DMP builds as patch which transitions text2 to text1.

If you are sure that such swaps of lengths and starts really produce a valdi reverse patch, please add a link to docs here (or write your own).

SashaOrlova: ok, change

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,39 +18,23 @@ import java.util.LinkedList
...
+        val dmp = diff_match_patch()
+        val newText = dmp.patch_apply(patchList, text)
+        if ((newText[1] as BooleanArray).any { !it }) {
+            throw DeletePatchException("Failure in patch apply")
         }
-        return reversePatchList
+        return dmp.patch_make(newText[0] as String, text)
     }
 
     fun deletePatch(deletedPatch: LinkedList<diff_match_patch.Patch>,
```
dbarashev: This patch is not yet "deleted" it is "deleteCandidate"

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,39 +18,23 @@ import java.util.LinkedList
...
                     nextPatches: List<diff_match_patch.Patch>, text: String): LinkedList<diff_match_patch.Patch> {
         val dmp = diff_match_patch()
-        var textVersion = text
+        var textVersion = dmp.patch_apply(deletedPatch, text)[0] as String
         val applyList = LinkedList<diff_match_patch.Patch>()
```
dbarashev: Please move this variable into the loop, so that it was obvious that it is needed only there.

dbarashev (approved):
LGTM [2 comments]
Feel free to merge having fixed the remaining issues.

Issue #27, opened 2018-05-23
Add patch id and GRPC method to delete patch


dbarashev:
So now we have some code which can safely delete patches and we need a convenient API for it. 

When user rejects some patch in the browser client he just clicks a button on corresponding changes. After that we need to identify the patch which needs to be deleted and send request to Papeeria server and eventually to Cosmas. 

We thus need some way of patch identification and GRPC method "deletePatch" which will accept the patch identifier, delete patch appropriately and will return the resulting text.

Perhaps a pair "file_id, timestamp" or a triple "project_id, file_id, timestamp" is a good patch identifier provided that all timestamps for the given file are different.However if they are not then we can build patch id from file generation number and patch position or from user_id and MD5 hash of patch text.

PR #28, opened 2018-05-25
add opportunity delete patch from storage

SashaOrlova:
#27

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -156,43 +156,41 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
-    override fun getTextWithoutPatch(request: CosmasProto.getTextWithoutPatchRequest,
-                              responseObserver: StreamObserver<CosmasProto.getTextWithoutPatchResponse>) {
-        var askedVersion = request.getVersionRequest.version
-        val blobRequestVersion: Blob? = try {
-            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
-                    Storage.BlobGetOption.generationMatch(askedVersion))
+    private fun getPatchListAndPreviousText(fileId: String, timestamp: Long): Pair<List<CosmasProto.Patch>, String> {
+        val blobs: Page<Blob> = try {
+            this.storage.list(this.bucketName, Storage.BlobListOption.versions(true),
+                    Storage.BlobListOption.prefix(fileId))
         } catch (e: StorageException) {
-            handleStorageException(e, responseObserver)
-            return
+            throw e
         }
-        if (blobRequestVersion == null) {
-            val requestStatus = Status.NOT_FOUND.withDescription(
-                    "Can't delete patch. There is no such file version in storage")
-            responseObserver.onError(StatusException(requestStatus))
-            return
-        }
-        val blobPreviousVersion: Blob?  = try {
-            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
-                    Storage.BlobGetOption.generationMatch(askedVersion - 1))
-        } catch (e: StorageException) {
-            handleStorageException(e, responseObserver)
-            return
-        }
-        if (blobPreviousVersion == null && askedVersion > 0) {
-            val requestStatus = Status.NOT_FOUND.withDescription(
-                    "Can't delete patch. There is no such file version in storage")
-            responseObserver.onError(StatusException(requestStatus))
-            return
-        }
-        val fileVersion1 = CosmasProto.FileVersion.parseFrom(blobRequestVersion.getContent())
-        val fileVersion2 = if (blobPreviousVersion != null) { CosmasProto.FileVersion.parseFrom(blobPreviousVersion.getContent()) } else { null }
-        val text = if (fileVersion2 != null) { fileVersion2.content.toStringUtf8() } else {""}
         val patchList = mutableListOf<CosmasProto.Patch>()
-        patchList.addAll(fileVersion1.patchesList)
+        var closestTimestamp = -1L
+        var previousText = ""
+        blobs.iterateAll().forEach {
+            if (it.createTime >= timestamp) {
+                patchList.addAll(CosmasProto.FileVersion.parseFrom(it.getContent()).patchesList)
+            }
+            else if (it.createTime > closestTimestamp) {
+                closestTimestamp = it.createTime
+                previousText = CosmasProto.FileVersion.parseFrom(it.getContent()).content.toStringUtf8()
+            }
+        }
+        patchList.sortBy { patch -> patch.timestamp }
+        return Pair(patchList, previousText)
+    }
+
+
+    override fun deletePatch(request: CosmasProto.DeletePatchRequest,
+                             responseObserver: StreamObserver<CosmasProto.DeletePatchResponse>) {
+        val (patchList, text) = try {
+            getPatchListAndPreviousText(request.fileId, request.versionTimestamp)
+        } catch (e: StorageException) {
+            handleStorageException(e, responseObserver)
+            return
+        }
         var indexCandidateDeletePatch = -1
-        for ((patchIndex, patch) in patchList.withIndex()) {
-            if (patch.timeStamp == request.patchTimeStamp) {
+        for ((patchIndex, patch) in patchList.iterator().withIndex()) {
+            if (patch.timestamp == request.patchTimestamp) {
                 indexCandidateDeletePatch = patchIndex
             }
         }
@@ -202,28 +200,23 @@ class CosmasGoogleCloudService(private val bucketName: String,
             responseObserver.onError(StatusException(requestStatus))
             return
         }
-        while (true) {
-            askedVersion++
-            val blob: Blob = try {
-                this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
-                        Storage.BlobGetOption.generationMatch(askedVersion))
-            } catch (e: StorageException) {
-                handleStorageException(e, responseObserver)
-                return
-            } ?: break
-            val fileVersion = CosmasProto.FileVersion.parseFrom(blob.getContent())
-            patchList.addAll(fileVersion.patchesList)
-        }
+        val textNoPatch : String
+        try {
         val textBeforeCandidateDelete = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
         val finishText = PatchCorrector.applyPatch(patchList, text)
-        val textNoPatch = PatchCorrector.applyPatch(
+        textNoPatch = PatchCorrector.applyPatch(
                 PatchCorrector.deletePatch(
                         patchList[indexCandidateDeletePatch],
                         patchList.subList(indexCandidateDeletePatch + 1, patchList.size),
-                textBeforeCandidateDelete),
+                        textBeforeCandidateDelete),
                 finishText
-        )
-        val response = CosmasProto.getTextWithoutPatchResponse.newBuilder()
+            )
+        } catch (e : PatchCorrector.ApplyPatchException) {
+            val status = Status.ABORTED.withDescription(e.message)
+            responseObserver.onError(StatusException(status))
+            return
+        }
+        val response = CosmasProto.DeletePatchResponse.newBuilder()
         response.content = ByteString.copyFromUtf8(textNoPatch)
         responseObserver.onNext(response.build())
         responseObserver.onCompleted()
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -16,8 +16,10 @@ package com.bardsoftware.papeeria.backend.cosmas
 
 import com.google.api.gax.paging.Page
 import com.google.cloud.storage.*
+import com.google.protobuf.ByteString
 import io.grpc.Status
 import io.grpc.StatusException
+import io.grpc.internal.testing.StreamRecorder
 import io.grpc.stub.StreamObserver
 import org.slf4j.LoggerFactory
 import java.util.concurrent.ConcurrentHashMap
@@ -155,6 +157,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    fun deletePatchFromStorage(request: CosmasProto.DeletePatchRequest,
+                              responseObserver: StreamObserver<CosmasProto.DeletePatchResponse>) {
+        val blob: Blob? = try {
+            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
+                    Storage.BlobGetOption.generationMatch(request.getVersionRequest.version))
+        } catch (e: StorageException) {
+            handleStorageException(e, responseObserver)
+            return
+        }
+        val response = CosmasProto.DeletePatchResponse.newBuilder()
+        if (blob == null) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such file version in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val fileVersion = CosmasProto.FileVersion.parseFrom(blob.getContent())
+        val text = fileVersion.content.toStringUtf8()
+        val patchList = fileVersion.patchesList
+        var indexCandidateDeletePatch = -1
+        for (patch in patchList) {
+            if (patch.timeStamp == request.timeStamp) {
+                indexCandidateDeletePatch = patchList.indexOf(patch)
+            }
+        }
+        if (indexCandidateDeletePatch == -1) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such patch in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val newText = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
+        val textNoPatch = PatchCorrector.applyPatch(
+                PatchCorrector.deletePatch(
+                        PatchCorrector.textToPatch(patchList[indexCandidateDeletePatch]),
+                        PatchCorrector.textToPatches(patchList.subList(indexCandidateDeletePatch + 1, patchList.size)),
+                        newText
+                ),
+                PatchCorrector.applyPatch(patchList, text)
+        )
+        response.content = ByteString.copyFromUtf8(textNoPatch)
+        responseObserver.onNext(response.build())
+        responseObserver.onCompleted()
+    }
+
     private fun handleStorageException(e: StorageException, responseObserver: StreamObserver<*>) {
         LOG.error("StorageException happened: ${e.message}")
         responseObserver.onError(e)
@@ -182,4 +229,5 @@ class CosmasGoogleCloudService(private val bucketName: String,
         }
         return CosmasProto.FileVersion.parseFrom(blob?.getContent()).patchesList
     }
+
 }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -182,8 +182,11 @@ class CosmasGoogleCloudService(private val bucketName: String,
 
     override fun deletePatch(request: CosmasProto.DeletePatchRequest,
                              responseObserver: StreamObserver<CosmasProto.DeletePatchResponse>) {
+        // Timestamp of file version witch contains patch
+        val versionTimestamp = this.storage.get(BlobId.of(this.bucketName, request.fileId, request.generation)).createTime
+        // Text of version from which patch was applied (version before version witch contains patch)
         val (patchList, text) = try {
-            getPatchListAndPreviousText(request.fileId, request.versionTimestamp)
+            getPatchListAndPreviousText(request.fileId, versionTimestamp)
         } catch (e: StorageException) {
             handleStorageException(e, responseObserver)
             return
@@ -192,6 +195,7 @@ class CosmasGoogleCloudService(private val bucketName: String,
         for ((patchIndex, patch) in patchList.iterator().withIndex()) {
             if (patch.timestamp == request.patchTimestamp) {
                 indexCandidateDeletePatch = patchIndex
+                break
             }
         }
         if (indexCandidateDeletePatch == -1) {
@@ -200,24 +204,25 @@ class CosmasGoogleCloudService(private val bucketName: String,
             responseObserver.onError(StatusException(requestStatus))
             return
         }
-        val textNoPatch : String
+        val textWithoutPatch: String
         try {
-        val textBeforeCandidateDelete = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
-        val finishText = PatchCorrector.applyPatch(patchList, text)
-        textNoPatch = PatchCorrector.applyPatch(
-                PatchCorrector.deletePatch(
-                        patchList[indexCandidateDeletePatch],
-                        patchList.subList(indexCandidateDeletePatch + 1, patchList.size),
-                        textBeforeCandidateDelete),
-                finishText
-            )
-        } catch (e : PatchCorrector.ApplyPatchException) {
-            val status = Status.ABORTED.withDescription(e.message)
+            val textBeforeCandidateDelete = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
+            val finishText = CosmasProto.FileVersion.parseFrom(
+                    this.storage.get(BlobId.of(this.bucketName, request.fileId)).getContent()).content.toStringUtf8()
+            textWithoutPatch = PatchCorrector.applyPatch(
+                    PatchCorrector.deletePatch(
+                            patchList[indexCandidateDeletePatch],
+                            patchList.subList(indexCandidateDeletePatch + 1, patchList.size),
+                            textBeforeCandidateDelete),
+                    finishText)
+        } catch (e: PatchCorrector.ApplyPatchException) {
+            val status = Status.INTERNAL.withDescription(e.message)
+            LOG.error("Can't apply patch: ${e.message}")
             responseObserver.onError(StatusException(status))
             return
         }
         val response = CosmasProto.DeletePatchResponse.newBuilder()
-        response.content = ByteString.copyFromUtf8(textNoPatch)
+        response.content = ByteString.copyFromUtf8(textWithoutPatch)
         responseObserver.onNext(response.build())
         responseObserver.onCompleted()
     }
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -19,7 +19,6 @@ import com.google.cloud.storage.*
 import com.google.protobuf.ByteString
 import io.grpc.Status
 import io.grpc.StatusException
-import io.grpc.internal.testing.StreamRecorder
 import io.grpc.stub.StreamObserver
 import org.slf4j.LoggerFactory
 import java.util.concurrent.ConcurrentHashMap
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -156,29 +156,44 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
-    fun deletePatchFromStorage(request: CosmasProto.DeletePatchRequest,
-                              responseObserver: StreamObserver<CosmasProto.DeletePatchResponse>) {
-        val blob: Blob? = try {
+    override fun getTextWithoutPatch(request: CosmasProto.getTextWithoutPatchRequest,
+                              responseObserver: StreamObserver<CosmasProto.getTextWithoutPatchResponse>) {
+        var askedVersion = request.getVersionRequest.version
+        val blobRequestVersion: Blob? = try {
             this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    fun deletePatchFromStorage(request: CosmasProto.DeletePatchRequest,
```
dbarashev: You know, this method **does not** delete patch _from storage_. It builds a new text with the given patch excluded. 

Also, did you mean to implement the service method? If yes then you'll probably want to name the function exactly like it is named in the superclass (`deletePatch`) and add `override` keyword.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+                              responseObserver: StreamObserver<CosmasProto.DeletePatchResponse>) {
+        val blob: Blob? = try {
+            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
+                    Storage.BlobGetOption.generationMatch(request.getVersionRequest.version))
+        } catch (e: StorageException) {
+            handleStorageException(e, responseObserver)
+            return
+        }
+        val response = CosmasProto.DeletePatchResponse.newBuilder()
```
dbarashev: The first usage of this val is 30 lines below. Please move it closer to the usage.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val fileVersion = CosmasProto.FileVersion.parseFrom(blob.getContent())
+        val text = fileVersion.content.toStringUtf8()
+        val patchList = fileVersion.patchesList
+        var indexCandidateDeletePatch = -1
+        for (patch in patchList) {
+            if (patch.timeStamp == request.timeStamp) {
+                indexCandidateDeletePatch = patchList.indexOf(patch)
```
dbarashev: There is nice [`withIndex`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/with-index.html) library function in Kotlin libs

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            }
+        }
+        if (indexCandidateDeletePatch == -1) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such patch in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val newText = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
```
dbarashev: Is my understanding correct that in a single stored `FileVersion` we have some text with version N and patches which were applied _before_ we created version N, that is, from version N-1 to N? If so then what's the point of applying patches to `text` ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+        val textNoPatch = PatchCorrector.applyPatch(
```
dbarashev: This piece of code makes me thinking that `PatchCorrector` could have had better interface which would allow the caller to just pass a list of patches and identifier of one which has to be removed.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,51 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+                PatchCorrector.deletePatch(
+                        PatchCorrector.textToPatch(patchList[indexCandidateDeletePatch]),
+                        PatchCorrector.textToPatches(patchList.subList(indexCandidateDeletePatch + 1, patchList.size)),
+                        newText
+                ),
+                PatchCorrector.applyPatch(patchList, text)
+        )
+        response.content = ByteString.copyFromUtf8(textNoPatch)
+        responseObserver.onNext(response.build())
+        responseObserver.onCompleted()
```
dbarashev: What shall we do if there are other versions of file? That is, when our current file version in N and client (==user) directed to remove a patch between versions N-6 and N-5, he obviously wants to get the result of changing the current version, not version N-5.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,9 +18,35 @@ import java.util.LinkedList
 import name.fraser.neil.plaintext.diff_match_patch
 
 object PatchCorrector {
+    private val dmp = diff_match_patch()
+
+    fun textToPatches(patchList: MutableList<CosmasProto.Patch>) : LinkedList<diff_match_patch.Patch> {
+        val linkedList = LinkedList<diff_match_patch.Patch>()
+        for (patch in patchList)
```
dbarashev: Please. Always. Surround. With. Curly. Brackets.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,9 +18,35 @@ import java.util.LinkedList
...
+            linkedList.addAll(textToPatch(patch))
+        return linkedList
+    }
+
+    fun textToPatch(patch : CosmasProto.Patch) : LinkedList<diff_match_patch.Patch> {
+        val patchList = dmp.patch_fromText(patch.text)
+        val linkedList = LinkedList<diff_match_patch.Patch>()
```
dbarashev: Why don't you just return what `patch_fromText` returns? What's the point of this additional list?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,9 +18,35 @@ import java.util.LinkedList
...
+        linkedList.addAll(patchList)
+        println(patchList[0])
+        return linkedList
+    }
+
+    fun applyPatch(patchList: LinkedList<diff_match_patch.Patch>, text: String) : String {
```
dbarashev: Why do you need a method which doesn't check for errors?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,9 +18,35 @@ import java.util.LinkedList
 import name.fraser.neil.plaintext.diff_match_patch
 
 object PatchCorrector {
+    private val dmp = diff_match_patch()
+
+    fun textToPatches(patchList: MutableList<CosmasProto.Patch>) : LinkedList<diff_match_patch.Patch> {
```
dbarashev: When you require "MutableList" in your function, you want to say to the external world that your intention is to change the list contents. If you don't plan to change the list, require immutable one (which is just `List` in Kotlin collection interfaces)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,9 +18,35 @@ import java.util.LinkedList
...
+        return linkedList
+    }
+
+    fun applyPatch(patchList: LinkedList<diff_match_patch.Patch>, text: String) : String {
+        return dmp.patch_apply(patchList, text)[0] as String
+    }
+
+    fun applyPatch(patchList: MutableList<CosmasProto.Patch>, text: String) : String {
+        val newText = dmp.patch_apply(textToPatches(patchList), text)
+        if ((newText[1] as BooleanArray).any { !it })
```
dbarashev: Curly braces.

src/main/proto/cosmas.proto
```diff
@@ -80,6 +80,15 @@ message FileVersion {
     bytes content = 2;
 }
 
+message DeletePatchRequest {
+    GetVersionRequest getVersionRequest = 1;
+    int64 timeStamp = 2;
```
dbarashev: timeStamp refers pretty much to anything, in particular, there is timestamp in `FileVersionInfo`. I suggest to make it more clear, e.g. `patchTimeStamp`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I'll rename function and add in superclass

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok, I think that it text and patches after. Change

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok, it wasn't obvious

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: because I want it returns LinkedList, not MutableList

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: change

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: ok

src/main/proto/cosmas.proto
SashaOrlova: ok

src/main/proto/cosmas.proto
```diff
@@ -80,6 +80,15 @@ message FileVersion {
     bytes content = 2;
 }
 
+message getTextWithoutPatchRequest {
```
dbarashev: Messages are equivalent of classes. Please capitalize their names.

src/main/proto/cosmas.proto
```diff
@@ -91,4 +100,6 @@ service Cosmas {
     }
     rpc CommitVersion (CommitVersionRequest) returns (CommitVersionResponse) {
     }
+    rpc getTextWithoutPatch (getTextWithoutPatchRequest) returns (getTextWithoutPatchResponse) {
```
dbarashev: This "getTextWithoutPatch" looks clumsy. I was happy with "deletePatch", "removePatch", or anything like this. My comment was about _the storage_ part of the method name: you delete patch, but not from the storage.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    override fun getTextWithoutPatch(request: CosmasProto.getTextWithoutPatchRequest,
+                              responseObserver: StreamObserver<CosmasProto.getTextWithoutPatchResponse>) {
```
dbarashev: Please align arguments properly.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+        }
+        if (blobRequestVersion == null) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such file version in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val blobPreviousVersion: Blob?  = try {
+            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
+                    Storage.BlobGetOption.generationMatch(askedVersion - 1))
```
dbarashev: Why do you think that previous version has generation which is 1 less than current? GCS provides no such promise. Generation is pretty much like timestamp.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            handleStorageException(e, responseObserver)
+            return
+        }
+        if (blobRequestVersion == null) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such file version in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val blobPreviousVersion: Blob?  = try {
```
dbarashev: Please remove one space before `=`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            return
+        }
+        if (blobPreviousVersion == null && askedVersion > 0) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such file version in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        val fileVersion1 = CosmasProto.FileVersion.parseFrom(blobRequestVersion.getContent())
+        val fileVersion2 = if (blobPreviousVersion != null) { CosmasProto.FileVersion.parseFrom(blobPreviousVersion.getContent()) } else { null }
```
dbarashev: The only use of this val is one line below where you check if it is null or not. Maybe just inline `.content.toStringUtf8()` call into this statement?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            }
+        }
+        if (indexCandidateDeletePatch == -1) {
+            val requestStatus = Status.NOT_FOUND.withDescription(
+                    "Can't delete patch. There is no such patch in storage")
+            responseObserver.onError(StatusException(requestStatus))
+            return
+        }
+        while (true) {
+            askedVersion++
```
dbarashev: This. Will. Not. Work.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            val blob: Blob = try {
+                this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
+                        Storage.BlobGetOption.generationMatch(askedVersion))
+            } catch (e: StorageException) {
+                handleStorageException(e, responseObserver)
+                return
+            } ?: break
```
dbarashev: With all due respect to Kotlin syntax, this reads ugly. However, this will likely vanish when you change the logic of iterating through the generations.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+            val fileVersion = CosmasProto.FileVersion.parseFrom(blob.getContent())
+            patchList.addAll(fileVersion.patchesList)
+        }
+        val textBeforeCandidateDelete = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
+        val finishText = PatchCorrector.applyPatch(patchList, text)
```
dbarashev: Finish text is the contents of the latest version, no?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -155,6 +156,79 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    override fun getTextWithoutPatch(request: CosmasProto.getTextWithoutPatchRequest,
```
dbarashev: This method is pretty long and I suggest extracting separate functions responsible for:

1. Get an ordered list of all required `FileVersion` objects (rough equivalent of lines 160-190)
2. Build a list of patches from the list of `FileVersion` objects (rough equivalent of lines 191-216)
3. Build the result (the remaining lines)

If you give these functions meaningful names, the code of this method will be self-documenting:
```
getVersionTail
buildPatchListTail
removePatch
```

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -18,19 +18,54 @@ import java.util.LinkedList
 import name.fraser.neil.plaintext.diff_match_patch
 
 object PatchCorrector {
+    private val dmp = diff_match_patch()
+
+    private fun textToPatches(patchList: List<CosmasProto.Patch>) : LinkedList<diff_match_patch.Patch> {
+        val newPatchList = LinkedList<diff_match_patch.Patch>()
+        for (patch in patchList) {
+            newPatchList.addAll(textToPatch(patch))
```
dbarashev: Can't help thinking that replacing `textToPatch(patch)` with `dmp.patch_fromText(patch.text)` will make this code more clear and efficient. Perhaps the other usage of `textToPatch` can be replaced with the same call and that method will no longer be needed?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
```diff
@@ -49,4 +84,5 @@ object PatchCorrector {
     }
 
     public class DeletePatchException(message: String) : Throwable(message)
+    public class ApplyPatchException(message: String) : Throwable(message)
```
dbarashev: You throw this exception but never catch it, so from the usage perspective it is no different than just RuntimeException or NullPointerException or whatever. The goal of introducing your own class must be ability to catch and handle exceptions of that class.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: I want this method to return LinkedList, not MutableList, for use it with dmp without conversion MutableList to LinkedList

src/main/proto/cosmas.proto
SashaOrlova: ok

src/main/proto/cosmas.proto
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: ok, change to timestamp for FileVersion

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I delete this code

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I delete this code

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: change

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I delete this code

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: yes, but what difference how I get it?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
SashaOrlova: I divide function into two parts: get patchList and take version without patch, I don't think, that I need more pieces

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
SashaOrlova: add exception handling

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/PatchCorrector.kt
dbarashev: `LinkedList(dmp.patch_fromText(deleteCandidate.text))`

SashaOrlova: delete function

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -156,43 +156,41 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
-            this.storage.get(BlobInfo.newBuilder(this.bucketName, request.getVersionRequest.fileId).build().blobId,
-                    Storage.BlobGetOption.generationMatch(askedVersion))
+    private fun getPatchListAndPreviousText(fileId: String, timestamp: Long): Pair<List<CosmasProto.Patch>, String> {
+        val blobs: Page<Blob> = try {
+            this.storage.list(this.bucketName, Storage.BlobListOption.versions(true),
+                    Storage.BlobListOption.prefix(fileId))
         } catch (e: StorageException) {
-            handleStorageException(e, responseObserver)
-            return
+            throw e
```
dbarashev: If re-throwing is the only thing which you plan to do then you can probably just do not catch it.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -156,43 +156,41 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
-        if (blobRequestVersion == null) {
-            val requestStatus = Status.NOT_FOUND.withDescription(
-                    "Can't delete patch. There is no such file version in storage")
-            responseObserver.onError(StatusException(requestStatus))
-            return
+        val patchList = mutableListOf<CosmasProto.Patch>()
+        var closestTimestamp = -1L
+        var previousText = ""
+        blobs.iterateAll().forEach {
+            if (it.createTime >= timestamp) {
```
dbarashev, line 170: From GCS point, version is identified by [`generation number`](https://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/index.html). I suggest using generation for searching for the "anchor" blob. Generation is available to the browser client so it is okay to require it in the interface.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -156,43 +156,41 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
-        val fileVersion2 = if (blobPreviousVersion != null) { CosmasProto.FileVersion.parseFrom(blobPreviousVersion.getContent()) } else { null }
-        val text = if (fileVersion2 != null) { fileVersion2.content.toStringUtf8() } else {""}
-        val patchList = mutableListOf<CosmasProto.Patch>()
-        patchList.addAll(fileVersion1.patchesList)
         var indexCandidateDeletePatch = -1
-        for ((patchIndex, patch) in patchList.withIndex()) {
-            if (patch.timeStamp == request.patchTimeStamp) {
+        for ((patchIndex, patch) in patchList.iterator().withIndex()) {
+            if (patch.timestamp == request.patchTimestamp) {
                 indexCandidateDeletePatch = patchIndex
```
dbarashev: If you have found the patch with matching timestamp, you don't need to iterate any longer. Break here.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
dbarashev: You don't burn CPU cycles on applying patches to get what is already available to you in the list of all object versions.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -202,28 +200,23 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
                         patchList[indexCandidateDeletePatch],
                         patchList.subList(indexCandidateDeletePatch + 1, patchList.size),
-                textBeforeCandidateDelete),
+                        textBeforeCandidateDelete),
                 finishText
-        )
-        val response = CosmasProto.getTextWithoutPatchResponse.newBuilder()
+            )
+        } catch (e : PatchCorrector.ApplyPatchException) {
+            val status = Status.ABORTED.withDescription(e.message)
```
dbarashev: ABORTED has very specific meaning and I suggest using just INTERNAL.
Besides, you really want to log the exception here to error log. Otherwise you will have no even minimal debug information in production.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -202,28 +200,23 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
-            } catch (e: StorageException) {
-                handleStorageException(e, responseObserver)
-                return
-            } ?: break
-            val fileVersion = CosmasProto.FileVersion.parseFrom(blob.getContent())
-            patchList.addAll(fileVersion.patchesList)
-        }
+        val textNoPatch : String
+        try {
         val textBeforeCandidateDelete = PatchCorrector.applyPatch(patchList.subList(0, indexCandidateDeletePatch), text)
```
dbarashev: Something is wrong with indents here.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: fixed

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

dbarashev (approved):
LGTM

Issue #35, opened 2018-06-27
Verify validity of patch list


dbarashev:
We store subsequent file versions and list of patches between them. We assume that if we apply patches to version N then we get version N+1. However, we do not check if it is really true. 

We have a couple of options:
1. Validate the sequence of patches on e.g. commit requests and check that version N + patches = version N+1
1. Send only patches from Papeeria FE to Cosmas and re-construct version N+1 from version N+accumulated patches on commit request.

I think we should proceed with option 2, because it reduces the amount of data exchanged between FE and Cosmas (we send only patches, not the actual file contents). At the same time, we need to apply the accumulated list of patches to version N in any scenario, be it validation or building of version N+1

PR #36, opened 2018-06-27
Change createPatch fun and change some tests

iisuslik43:
#35

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -22,6 +22,7 @@ import com.google.protobuf.ByteString
 import io.grpc.Status
 import io.grpc.StatusException
 import io.grpc.stub.StreamObserver
+import name.fraser.neil.plaintext.diff_match_patch
 import org.slf4j.LoggerFactory
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ConcurrentMap
@@ -319,6 +320,49 @@ class CosmasGoogleCloudService(private val bucketName: String,
         responseObserver.onCompleted()
     }
 
+    override fun forcedFileCommit(request: CosmasProto.ForcedFileCommitRequest,
+                                  responseObserver: StreamObserver<CosmasProto.ForcedFileCommitResponse>) {
+        LOG.info("Get request for commit file # ${request.fileId} in force")
+        val project = synchronized(this.fileBuffer) {
+            this.fileBuffer.getValue(request.projectId)
+        }
+        val potentialVersion = synchronized(project) {
+            project[request.fileId]
+        }
+        val previousVersion = potentialVersion?.content?.toStringUtf8() ?: ""
+        val actualVersion = request.actualContent.toStringUtf8()
+        val diffPatch = diff_match_patch().patch_toText(diff_match_patch().patch_make(previousVersion, actualVersion))
+        val patch = CosmasProto.Patch.newBuilder()
+                .setText(diffPatch)
+                .setUserId("0")
+                .setTimestamp(System.currentTimeMillis())
+                .setActualHash(md5Hash(actualVersion))
+                .build()
+        val newVersion = CosmasProto.FileVersion.newBuilder()
+                .addPatches(patch)
+                .setContent(ByteString.copyFrom(actualVersion.toByteArray()))
+                .build()
+        try {
+            this.storage.create(
+                    BlobInfo.newBuilder(this.bucketName, request.fileId).build(),
+                    newVersion.toByteArray())
+        } catch (e: StorageException) {
+            handleStorageException(e, responseObserver)
+            return
+        }
+        synchronized(project) {
+            project[request.fileId] = newVersion.toBuilder()
+                    .clearPatches()
+                    .build()
+        }
+        synchronized(this.fileBuffer) {
+            this.fileBuffer[request.projectId] = project
+        }
+        val response = CosmasProto.ForcedFileCommitResponse.newBuilder().build()
+        responseObserver.onNext(response)
+        responseObserver.onCompleted()
+    }
+
     private fun handleStorageException(e: StorageException, responseObserver: StreamObserver<*>) {
         LOG.error("StorageException happened: ${e.message}")
         responseObserver.onError(e)
@@ -335,10 +379,6 @@ class CosmasGoogleCloudService(private val bucketName: String,
 
     fun getPatchList(projectId: String, fileId: String): List<CosmasProto.Patch>? {
         return fileBuffer[projectId]?.get(fileId)?.patchesList
-    }
-
-    fun getFileFromBuffer(projectId: String, fileId: String): String? {
-        return fileBuffer[projectId]?.get(fileId)?.content?.toStringUtf8()
     }
 
     fun getPatchListFromStorage(fileId: String, version: Long): List<CosmasProto.Patch>? {
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -63,16 +63,10 @@ class CosmasGoogleCloudService(private val bucketName: String,
             val project = this.fileBuffer.getValue(request.projectId)
             val fileVersion = project[request.fileId]
             if (fileVersion != null) {
-                val text = fileVersion.content.toStringUtf8()
-                val newText = PatchCorrector.applyPatch(listOf(request.patch), text)
-                project[request.fileId] = fileVersion.toBuilder().
-                        addPatches(request.patch).
-                        setContent(ByteString.copyFrom(newText.toByteArray()))
-                        .build()
+                project[request.fileId] = fileVersion.toBuilder().addPatches(request.patch).build()
             } else {
-                val newText = PatchCorrector.applyPatch(listOf(request.patch), "")
                 project[request.fileId] = CosmasProto.FileVersion.newBuilder().
-                        setContent(ByteString.copyFrom(newText.toByteArray())).
+                        setContent(ByteString.copyFrom("".toByteArray())).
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -65,7 +65,10 @@ class CosmasGoogleCloudService(private val bucketName: String,
             if (fileVersion != null) {
                 project[request.fileId] = fileVersion.toBuilder().addPatches(request.patch).build()
             } else {
-                project[request.fileId] = CosmasProto.FileVersion.newBuilder().addPatches(request.patch).build()
+                project[request.fileId] = CosmasProto.FileVersion.newBuilder().
+                        setContent(ByteString.copyFrom("".toByteArray())).
```
dbarashev: `ByteString.EMPTY` ? Isn't it the default by any chance?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -234,11 +243,7 @@ class CosmasGoogleCloudService(private val bucketName: String,
             handleStorageException(e, responseObserver)
             return
         }
-        val newTomb = CosmasProto.FileTomb.newBuilder().
-                setFileId(request.fileId).
-                setFileName(request.fileName).
-                setRemovalTimestamp(request.removalTimestamp).
-                build()
+        val newTomb = CosmasProto.FileTomb.newBuilder().setFileId(request.fileId).setFileName(request.fileName).setRemovalTimestamp(request.removalTimestamp).build()
```
dbarashev: Please revert to formatting with linebreaks before every setter call, it is way more readable. This applies through the whole pull request.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -295,6 +300,10 @@ class CosmasGoogleCloudService(private val bucketName: String,
         return fileBuffer[projectId]?.get(fileId)?.patchesList
     }
 
+    fun getFileFromBuffer(projectId: String, fileId: String): String? {
```
dbarashev: IDEA indicates that this is unused.

dbarashev (commented):
You probably need to catch ApplyPatchException and report the failure in the response, so that the client could take fallback measures (e.g. send new version as text and commit it)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I hope I fixed it everywhere

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, it's `ByteString.EMPTY` by default

src/main/proto/cosmas.proto
```diff
@@ -23,6 +23,11 @@ message FileVersionInfo {
     int64 timestamp = 2;
 }
 
+message FileInfo{
```
dbarashev: nit: space before '{'

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -319,6 +320,49 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+        }
+        val potentialVersion = synchronized(project) {
+            project[request.fileId]
+        }
+        val previousVersion = potentialVersion?.content?.toStringUtf8() ?: ""
+        val actualVersion = request.actualContent.toStringUtf8()
+        val diffPatch = diff_match_patch().patch_toText(diff_match_patch().patch_make(previousVersion, actualVersion))
+        val patch = CosmasProto.Patch.newBuilder()
+                .setText(diffPatch)
+                .setUserId("0")
```
dbarashev: Why not using the real user id here?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -298,6 +287,44 @@ class CosmasGoogleCloudServiceTest {
...
+        val patch = diffPatch(USER_ID, "", "ver1", 1)
+        addPatchToService(patch)
+        commit()
+        val uselessPatch = diffPatch(USER_ID, "lol", "kek", 2)
+        addPatchToService(uselessPatch)
+        forcedCommit("ver2")
+        //val diffPatch = diffPatch("0", "ver1", "ver2", System.currentTimeMillis())
+        val ver1 = createFileVersion("ver1").toBuilder().addAllPatches(listOf(patch)).build()
+        //val ver2 = createFileVersion("ver2").toBuilder().addAllPatches(listOf(diffPatch)).build()
+        verify(fakeStorage).create(any(BlobInfo::class.java),
```
dbarashev: Any sense to verify also that ver1 => (forced) ver2 patch was added?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -86,22 +100,49 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+                        val newVersion = fileVersion.toBuilder()
+                                .setContent(ByteString.copyFrom(newText.toByteArray()))
+                        this.storage.create(
+                                BlobInfo.newBuilder(this.bucketName, fileId).build(),
+                                newVersion.build().toByteArray())
+                        project[fileId] = newVersion
+                                .clearPatches()
+                                .build()
+                    } else {
+                        val badFile = CosmasProto.FileInfo.newBuilder()
```
dbarashev: Log failure here? File/project id and hash values (not that they are useful when looking at the logs, but may turn out useful in case of deeper investigations)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -86,22 +100,49 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+                                .setFileId(fileId)
+                                .setProjectId(request.projectId)
+                                .build()
+                        response.addBadFiles(badFile)
+                    }
+                } catch (e: PatchCorrector.ApplyPatchException) {
+                    val badFile = CosmasProto.FileInfo.newBuilder()
```
dbarashev: And here as well

dbarashev (commented):
Awesome work! 
I suggest adding tests for validation failures, e.g. check that bad files are not empty in case of hash mismatch

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: The problem is, that I set timestamp to `System.currentTimeMillis()` in this big patch, maybe I can use another thing that can tell me what tim is it, and mock it.

dbarashev: Perhaps it makes sense to supply timestamp in the request.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: Something like "COSMAS"? Because there can't be any user who decided to do it

dbarashev: Technically, there is a user who hits Compile or whose patch was the last one before auto-compile, but probably it is even better if we see that it is not actually a human who made this patch. 
What about `robot:::cosmas`, similar to our auto-compiler user name?

src/main/proto/cosmas.proto
iisuslik43: done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: ok, done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -123,18 +124,26 @@ class CosmasGoogleCloudService(private val bucketName: String,
                                 .clearPatches()
                                 .build()
                     } else {
+                        LOG.error("File # $fileId version in Cosmas not equals to version in Papeeria")
```
dbarashev: Log hash values:
"Commit failure: hash mismatch. File=$fileId  Cosmas hash=....  last actual hash=..."

dbarashev (approved):
LGTM [1 comment]

PR #39, opened 2018-07-18
Send last version to -1 in request and add deleting file from cemetery

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -161,8 +161,9 @@ class CosmasGoogleCloudService(private val bucketName: String,
     override fun getVersion(request: CosmasProto.GetVersionRequest,
                             responseObserver: StreamObserver<CosmasProto.GetVersionResponse>) {
         LOG.info("Get request for version ${request.version} file # ${request.fileId}")
+        val version = if (request.version == -1L) null else request.version
```
dbarashev: This means "gimme the last version"? Can you please add a comment to the version field in proto file and here about this special meaning of -1 and the semantics of other values of "version"? I mean, they are supposed to be "generation" values, right? If yes then I  suggest renaming: we already have "generation" in the protobuffers, so let's call the same things with the same names.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -577,21 +577,74 @@ class CosmasGoogleCloudServiceTest {
...
+        val fakeStorage = mock(Storage::class.java)
+        val time = 1L
+        val cemeteryName = "$PROJECT_ID-cemetery"
+        val tomb = CosmasProto.FileTomb.newBuilder()
+                .setFileId(FILE_ID)
+                .setFileName("file")
+                .setRemovalTimestamp(time)
+                .build()
+        val cemetery = FileCemetery.newBuilder().addCemetery(tomb).build()
+        val emptyCemetery = FileCemetery.newBuilder().addAllCemetery(listOf()).build()
```
dbarashev: I thought repeated fields were empty lists by default, no?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -577,21 +577,73 @@ class CosmasGoogleCloudServiceTest {
...
         this.service = CosmasGoogleCloudService(this.BUCKET_NAME, fakeStorage)
-        val streamRecorder = deletedFileList(projectId)
+        val streamRecorder = deletedFileList(PROJECT_ID)
         assertNull(streamRecorder.error)
         Mockito.verify(fakeStorage).get(eq(BlobId.of(this.BUCKET_NAME, cemeteryName)))
         assertEquals(DeletedFileListResponse.newBuilder().addFiles(tomb).build(), streamRecorder.values[0])
     }
 
+    @Test
+    fun restoreDeletedFileTest() {
```
dbarashev, line 641: Can you also add a test for the case when there is no cemetery yet?

dbarashev (approved):
LGTM [more tests appreciated]

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: done

PR #40, opened 2018-07-20
Change Files Ids

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -415,8 +426,41 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+        for (change in request.changesList) {
+            prevIds[change.newFileId] = change.oldFileId
+            synchronized(project) {
+                if (project.contains(change.oldFileId)) {
+                    project[change.newFileId] = project[change.oldFileId]
+                    project.remove(change.oldFileId)
+                }
+            }
+        }
+        try {
```
dbarashev: I suggest making it more reliable: commit project on receiving ChangeFileIdRequest ( => you saved the history associated with all old ids), clear `project` ( => you do not need to remap ids, and new ids will start filling `project` from scratch) and make sure you write the mapping. The order of project commit/write map events doesn't seem to be important, but if you do write map first and then commit, holding the lock on `project` during all that process, it will be more reliable.

src/main/proto/cosmas.proto
```diff
@@ -21,6 +21,7 @@ option java_outer_classname = "CosmasProto";
 message FileVersionInfo {
     int64 generation = 1;
     int64 timestamp = 2;
+    string actualFileId = 3;
```
dbarashev: Do you really need this amendment? I think `fileId` is okay, short and matching `GetVersionRequest::fileId`. If you really want to amend "file id", I suggest "effective", in a sense similar to "effective user id" in Unix systems: id which is not your real one, but on whose behalf you take some action.

iisuslik43: ok, I used just `fileId`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -415,16 +426,61 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
     private fun handleStorageException(e: StorageException, responseObserver: StreamObserver<*>) {
         LOG.error("StorageException happened: ${e.message}")
         responseObserver.onError(e)
     }
 
+    private fun getPrevIds(projectId: String): MutableMap<String, String> {
+        val mapName = "${projectId}-fileIdMap"
+        val mapBytes: Blob? = this.storage.get(BlobId.of(this.bucketName, mapName))
+        return if (mapBytes == null) {
+            mutableMapOf()
```
dbarashev: Just `mapOf()` ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -415,16 +426,61 @@ class CosmasGoogleCloudService(private val bucketName: String,
...
+
+
     private fun handleStorageException(e: StorageException, responseObserver: StreamObserver<*>) {
         LOG.error("StorageException happened: ${e.message}")
         responseObserver.onError(e)
     }
 
+    private fun getPrevIds(projectId: String): MutableMap<String, String> {
+        val mapName = "${projectId}-fileIdMap"
+        val mapBytes: Blob? = this.storage.get(BlobId.of(this.bucketName, mapName))
```
dbarashev: This can probably be written with Elvis operator:
```
val mapBytes = this.storage.get(BlobId.of(this.bucketName, mapName)) ?: return mapOf()
```

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

dbarashev (approved):
LGTM

PR #46, opened 2018-08-13
Not committing with empty patch list, fix createVersion function

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -85,11 +85,11 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
             val project = this.fileBuffer.getValue(request.info.projectId)
-            val patchList = project[request.fileId]?.patchesList ?: mutableListOf()
-            project[request.fileId] = CosmasProto.FileVersion.newBuilder()
-                    .setContent(request.file)
-                    .addAllPatches(patchList)
-                    .build()
+            val newVersion = project[request.fileId]?.toBuilder() ?: CosmasProto.FileVersion.newBuilder()
+            val oldText = newVersion.content.toStringUtf8()
+            val newText = request.file.toStringUtf8()
+            newVersion.addPatches(getDiffPatch(oldText, newText, this.clock.millis()))
```
dbarashev, line 91: Any sense to pass timestamps from the outside as well? Because now patches and delete requests have external timestamps while createVersion uses its own. 
If you decide to supply timestamp externally, feel free to do it in a separate PR.

dbarashev (approved):
LGTM [1 comment]

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: Sounds reasonable, however Papeeria doesn't use this function. But I think I have enough time to fix it.

PR #48, opened 2018-08-21
Fixing delete patch function

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -254,11 +254,11 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         responseObserver.onCompleted()
     }
 
-    private fun getPatchListAndPreviousText(fileId: String, timestamp: Long,
+    private fun getPatchListAndPreviousText(fileName: String, timestamp: Long,
                                             info: ProjectInfo): Pair<List<CosmasProto.Patch>, String> {
         val blobs: Page<Blob> = try {
             this.storage.list(bucketName(info.isFreePlan), Storage.BlobListOption.versions(true),
-                    Storage.BlobListOption.prefix(fileId))
+                    Storage.BlobListOption.prefix(fileName))
```
dbarashev, line 261: Any sense to update tests so that they expected no just any BlobListOption? Meanwhile we introduce significant changes to the way of addressing objects in GCS and our unit tests pass both before and after code changes, without changes in tests themselves.

dbarashev (approved):
LGTM, but you may want to update tests.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: done

PR #49, opened 2018-08-22
Saving user name

iisuslik43:


src/main/proto/cosmas.proto
```diff
@@ -54,9 +54,10 @@ message FileVersion {
 
 message Patch {
     string userId = 1;
-    string text = 2;
-    int64 timestamp = 3;
-    string actualHash = 4;
+    string userName = 2;
```
dbarashev, line 57: This is not the issue now, when this feature is not yet launched, but just for the record and for the future: these numbers are not just sequential numbers, they are embedded into the serialized protocol buffer instances and are used when deserializing to understand what byte goes where. Natually, if "2" used to be "text" and now becomes "userName" then we're having troubles. 
After launch we will not be able to afford changing the numbers (unless we appropriately change all the stored data).

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -543,6 +543,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         return CosmasProto.Patch.newBuilder()
                 .setText(diffPatch)
                 .setUserId(COSMAS_ID)
+                .setUserName("COSMAS")
```
dbarashev: Let's use something like "Papeeria" just to have some understandable name if we happen to forget to update this later.

dbarashev (approved):
LGTM [1.5 comments]

src/main/proto/cosmas.proto
iisuslik43: ok, i ll remember it

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

Issue #50, opened 2018-10-15
Maintain ordered list of file versions 


dbarashev:
User story: we want to show the first N file versions fast, and load the remaining on request.

Technical issue : we can't rely on any ordering of blobs returned by `list` method. There is no guarantee that more recent versions come first or vice versa.

Possible solution: keep our own piece of information with version ordering. Perhaps we can piggyback on cemetery and fileidmapping objects.

PR #51, opened 2018-10-15
Maintain sorted list of file versions

iisuslik43:
#50 
I m writing more tests now

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -200,16 +200,16 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
                 .setFileId(fileId)
                 // For in-memory storage implementation resultBlob.generation == null,
                 // but in this case we don't care about generation value, so I set it to 1L
-                .setGeneration(generation.let { 1L })
+                .setGeneration(generation ?: 1L)
                 .setTimestamp(timestamp)
+                .build()
+        val fileIdVersionsInfo = FileIdVersionsInfo.newBuilder()
+                .addAllVersionInfo(fileIdVersionsInfoList)
+                .addVersionInfo(newFileVersionInfo)
                 .build()
         this.storage.create(
                 getBlobInfo("${info.projectId}-$fileId-fileIdVersionsInfo", info),
-                FileIdVersionsInfo.newBuilder()
-                        .addAllVersionInfo(fileIdVersionsInfoList)
-                        .addVersionInfo(newFileVersionInfo)
-                        .build()
-                        .toByteArray())
+                fileIdVersionsInfo.toByteArray())
     }
 
     override fun getVersion(request: CosmasProto.GetVersionRequest,
@@ -262,7 +262,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
                 return
             }
 
-            versionList.addAll(fileIdVersionsInfo)
+            versionList.addAll(fileIdVersionsInfo.reversed())
             curFileId = prevIds[curFileId]
         }
         response.addAllVersions(versionList)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -181,6 +188,31 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+        return FileIdVersionsInfo.parseFrom(infoBytes.getContent()).versionInfoList
+    }
+
+    private fun updateFileIdVersionsInfo(info: ProjectInfo, fileId: String, generation: Long?, timestamp: Long) {
+        val fileIdVersionsInfoList = getFileIdVersionsInfo(info, fileId)
+        val newFileVersionInfo = FileVersionInfo.newBuilder()
+                .setFileId(fileId)
+                // For in-memory storage implementation resultBlob.generation == null,
+                // but in this case we don't care about generation value, so I set it to 1L
+                .setGeneration(generation ?: 1L)
```
dbarashev, line 204: Thanks for the comment. It probably makes sense to extract storage interface. We currently us a number of methods, such as "create regular object", "create cemetery", "create id mapping", "create version list", etc. If we extract an interface for them, we'll be able to store these different pieces of data in different places, if we need to. And yes, we'll be able to remove Elvis operator.

This is probably out of scope of this PR though, so feel free to just add TODO

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -181,6 +188,31 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+                .setTimestamp(timestamp)
+                .build()
+        val fileIdVersionsInfo = FileIdVersionsInfo.newBuilder()
+                .addAllVersionInfo(fileIdVersionsInfoList)
+                .addVersionInfo(newFileVersionInfo)
+                .build()
+        this.storage.create(
```
dbarashev, line 211: We need to cleanup this metadata as object versions go away.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -222,26 +254,28 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
-                val versionInfo = CosmasProto.FileVersionInfo.newBuilder()
-                        .setGeneration(it.generation)
-                        .setTimestamp(fileVersion.timestamp)
-                        .setFileId(curFileId)
-                versionList.add(versionInfo.build())
+            if (rest == -1) {
+                versionList.addAll(fileIdVersionsInfo.reversed())
+            } else {
+                val delta = min(rest, fileIdVersionsInfo.size)
+                versionList.addAll(fileIdVersionsInfo.reversed()
```
dbarashev, line 273: Why don't you reverse the list at the very beginning? Or even better, why not store it already reversed?

src/main/proto/cosmas.proto
```diff
@@ -65,6 +65,10 @@ message ChangeId {
     string newFileId = 2;
 }
 
+message FileIdVersionsInfo {
```
dbarashev, line 68: Can you rename it to FileHistoryMetadata ?

src/main/proto/cosmas.proto
```diff
@@ -92,6 +96,7 @@ message CreatePatchResponse {
 message FileVersionListRequest {
     ProjectInfo info = 1;
     string fileId = 2;
+    int32 count = 3; // if count is 0 or -1 response will contain all versions
```
dbarashev, line 99: So the bigger is the count, the slower and bigger is the response. I suggest adding offset, so that pagination was more efficient.

dbarashev (commented):
Cool tests!

dbarashev:
This can be discarded now?

PR #52, opened 2018-10-29
Fast way to show first N file versions

iisuslik43:
#50 
I created a new branch for this implementation cause my previous try is very different

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -58,7 +58,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
             ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileVersion>>()
                     .withDefault { ConcurrentHashMap() }
 
-    private val gsutilCommand = if (gsutilImageName != "") "docker run ${gsutilImageName} gsutil" else "gsutil"
+    private val gsutilCommand = if (gsutilImageName != "") "docker run $gsutilImageName gsutil" else "gsutil"
 
 
     companion object {
@@ -66,7 +66,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
             return Hashing.md5().newHasher().putString(text, Charsets.UTF_8).hash().toString()
         }
 
-        val COSMAS_ID = "robot:::cosmas"
+        const val COSMAS_ID = "robot:::cosmas"
 
         fun getNewWindow(newInfo: FileVersionInfo, oldWindow: List<FileVersionInfo>,
                          windowMaxSize: Int): MutableList<FileVersionInfo> {
@@ -287,7 +287,8 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
                 ?: emptyList()
     }
 
-    private fun getFileVersionListFromStorage(projectInfo: ProjectInfo, fileId: String, startGeneration: Long): List<FileVersionInfo> {
+    private fun getFileVersionListFromStorage(projectInfo: ProjectInfo, fileId: String,
+                                              startGeneration: Long): List<FileVersionInfo> {
 
         val blob: Blob? = this.storage.get(getBlobId(fileId, projectInfo, startGeneration)) // throws StorageException
         return if (blob != null) {
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -498,7 +498,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
                 .setTimestamp(curTime)
                 .build()
 
-        val newWindow = getNewWindow(storageVersionInfo, latestVersion?.historyWindowList ?: emptyList())
+        val newWindow = getNewWindow(storageVersionInfo, versionToCommit.historyWindowList)
         synchronized(project) {
             project[request.fileId] = versionToCommit.toBuilder()
                     .clearPatches()
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -56,7 +56,6 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
 
     private val fileBuffer =
             ConcurrentHashMap<String, ConcurrentMap<String, CosmasProto.FileVersion>>()
-                    .withDefault { ConcurrentHashMap() }
 
     private val gsutilCommand = if (gsutilImageName != "") "docker run $gsutilImageName gsutil" else "gsutil"
 
@@ -100,12 +99,9 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         }
         val userId = request.patchesList.first().userId
         LOG.info("Get request for create new patch of file={} by user={}", request.fileId, userId)
-        synchronized(this.fileBuffer) {
-            if (this.fileBuffer[request.info.projectId] == null) {
-                this.fileBuffer[request.info.projectId] = ConcurrentHashMap()
-            }
+        val project = synchronized(this.fileBuffer) {
+            this.fileBuffer.getOrPut(request.info.projectId) { ConcurrentHashMap() }
         }
-        val project = this.fileBuffer.getValue(request.info.projectId)
         synchronized(project) {
             val fileVersion = project[request.fileId]
             if (fileVersion != null) {
@@ -201,7 +197,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         // Just committing buffered version with new content and current timestamp to GCS
         val curTime = clock.millis()
         val newVersion = fileVersion.toBuilder()
-                .setContent(ByteString.copyFrom(newText.toByteArray()))
+                .setContent(ByteString.copyFromUtf8(newText))
                 .setTimestamp(curTime)
         val resBlob = this.storage.create(getBlobInfo(fileId, projectInfo), newVersion.build().toByteArray())
 
@@ -435,12 +431,9 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
     override fun forcedFileCommit(request: CosmasProto.ForcedFileCommitRequest,
                                   responseObserver: StreamObserver<CosmasProto.ForcedFileCommitResponse>) {
         LOG.info("Get request for commit file={} in force", request.fileId)
-        synchronized(this.fileBuffer) {
-            if (this.fileBuffer[request.info.projectId] == null) {
-                this.fileBuffer[request.info.projectId] = ConcurrentHashMap()
-            }
+        val project = synchronized(this.fileBuffer) {
+            this.fileBuffer.getOrPut(request.info.projectId) { ConcurrentHashMap() }
         }
-        val project = this.fileBuffer.getValue(request.info.projectId)
         synchronized(project) {
 
             // Getting last version from storage or default instance if it doesn't exist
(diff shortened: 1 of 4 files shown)


iisuslik43:
I will fix forced committing and add more tests later

src/main/proto/cosmas.proto
```diff
@@ -50,6 +50,7 @@ message FileVersion {
     repeated Patch patches = 1;
     bytes content = 2;
     int64 timestamp = 3;
+    repeated FileVersionInfo historyWindow = 4;
```
dbarashev, line 53: I suggest adding intermediate protocol buffer 
```
message FileVersionInfoList {
    repeated FileVersionInfo versions = 1;
}
```

and changing type of `historyWindow` to `FileVersionInfoList`. This will make operations with the frame more natural. `historyWindow.addAllVersions(newWindow)` is more understandable than `addAllHistoryWindow(newWindow)`

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -214,34 +240,18 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
-                        .setFileId(curFileId)
-                versionList.add(versionInfo.build())
-            }
-            curFileId = prevIds[curFileId]
-        }
-        versionList.sortBy { it.timestamp }
+        val versionList = getFileVersionList(request.info, request.fileId,
+                request.firstWindowIndex, request.windowsCount)
         response.addAllVersions(versionList)
         if (response.versionsList.isEmpty()) {
```
dbarashev: swap this check with the previous line? It will be more natural to check for emptiness and only then write to the result.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -254,6 +264,55 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+            versionInfos.addAll(getFileVersionListFromStorage(info, versionInfos.last(),
+                    firstWindowIndex, windowsCount - 1))
+            return versionInfos
+        } else {
+            return getFileVersionListFromStorage(info, versionInfos.last(),
+                    firstWindowIndex - 1, windowsCount)
+        }
+    }
+
+    private fun getFileVersionListFromMemory(projectInfo: ProjectInfo, fileId: String): MutableList<FileVersionInfo> {
```
dbarashev: Returning immutable list from this method will make more clear that method has no side effects and is not supposed to have. Let the client do whatever he wants with the immutable result (e.g. create a copy of it)

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -57,12 +60,21 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
 
+
     companion object {
         fun md5Hash(text: String): String {
             return Hashing.md5().newHasher().putString(text, Charsets.UTF_8).hash().toString()
         }
 
         val COSMAS_ID = "robot:::cosmas"
+
+        fun getNewWindow(newInfo: FileVersionInfo, oldWindow: List<FileVersionInfo>,
```
dbarashev: `create` or `build` is better prefix than `get`. Because normally `get` gets something existing, rather than builds somethign new

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -181,6 +186,41 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         responseObserver.onCompleted()
     }
 
+    // We have to pass newText there cause we re applying patches to previous version in commit
```
dbarashev: My impression is that we have to pass text here just because we can't build it from any other source whe nwe're doing forcedCommit, no?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -181,6 +186,41 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+                .build()
+        val newWindow = getNewWindow(newInfo, fileVersion.historyWindowList)
+        project[fileId] = newVersion
+                .clearPatches()
+                .clearHistoryWindow()
+                .addAllHistoryWindow(newWindow)
+                .build()
+    }
+
+    private fun getNewWindow(newInfo: FileVersionInfo,
```
dbarashev: Maybe remove this method and replace usages with direct calls to companion object?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -391,41 +434,47 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         val project = synchronized(this.fileBuffer) {
```
dbarashev, line 434: What do you think of holding the lock on `fileBuffer` during the whole run of this method? It seems that we're going to have troubles if someone modifies the in-memory structure for the current project when this code is running, no?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -391,41 +434,47 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+            val latestVersionInfo = FileVersionInfo.newBuilder()
+                    .setFileId(request.fileId)
+                    // For in-memory storage implementation resultBlob.generation == null,
+                    // but in this case we don't care about generation value, so I set it to 1L
+                    .setGeneration(latestVersionBlob.generation ?: 1L)
+                    .setTimestamp(latestVersion.timestamp)
+                    .build()
+            getNewWindow(latestVersionInfo, latestVersion.historyWindowList)
+        } else emptyList<FileVersionInfo>()
+        // Content is taken from the latest versions
```
dbarashev: This comment is misleading, because "latestVersion" is already used for the latest stored version. We, however, do not take the contents from there.

src/main/proto/cosmas.proto
iisuslik43: I dont like it because I have to build `FileVersionInfoList` each time I want to change history window

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: hm, it's outdated, it's immutable now

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: fixed

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, I ve rewritten it

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, I thought I would call it more then 2 times when I created this method

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, it's a good idea, I should think about locking `fileBuffer` or `project` in other methods too

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: fixed

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -100,6 +100,11 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         }
         val userId = request.patchesList.first().userId
         LOG.info("Get request for create new patch of file={} by user={}", request.fileId, userId)
+        synchronized(this.fileBuffer) {
+            if (this.fileBuffer[request.info.projectId] == null) {
+                this.fileBuffer[request.info.projectId] = ConcurrentHashMap()
```
dbarashev: I see that `fileBuffer` is already configured with default value. This requires using `getValue` though, but maybe it worth doing?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -123,64 +139,91 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+    // before we re calling this method, so we dont want to do it again.
+    // In "forcedCommit" new text version is given in request.
+    private fun commitFromMemoryToGCS(projectInfo: ProjectInfo, fileId: String, newText: String) {
+        val project = this.fileBuffer[projectInfo.projectId] ?: return
+        val fileVersion = project[fileId] ?: return
+
+        // Just committing buffered version with new content and current timestamp to GCS
+        val curTime = clock.millis()
+        val newVersion = fileVersion.toBuilder()
+                .setContent(ByteString.copyFrom(newText.toByteArray()))
```
dbarashev: btw there is `ByteString.copeFromUtf8`: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/ByteString#copyFromUtf8-java.lang.String-

dbarashev (approved):
LGTM [2 comments]

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: Now I m using `getOrPut`, cause i want to init project as fast as possible: earlier we were doing it in the end of the method so we had to lock `fileBuffer` for all of it

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -520,13 +513,14 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
 
-        val project = this.fileBuffer.getValue(request.info.projectId)
-
-        synchronized(project) {
-            for (change in request.changesList) {
-                if (project.contains(change.oldFileId)) {
-                    project[change.newFileId] = project[change.oldFileId]
-                    project.remove(change.oldFileId)
+        val project = this.fileBuffer[request.info.projectId]
+        if (project != null) {
```
dbarashev, line 517: Is it possible that project is null? What shall we do in this case?

dbarashev (approved):
LGTM [1 comment]

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I think it is a normal case, for example if we have never got any patches from this project and then this request happens. We should do nothing with it and everything will be OK

dbarashev: What if we had some history before, but for whatever reason Cumulus restarted. Immediately after restart we ran git sync in Papeeria. Will we report id change here and do nothing?

PR #54, opened 2018-12-10
Restoring files from storage to buffer

iisuslik43:
Before doing smth in methods like `createPatch`, `getVersionList`, `forcedCommit` and `changeFileId`, I m restoring file to buffer from GCS if buffer doesn't contain such file

It's too early to review this code, it needs more tests

### DIFF
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -129,7 +129,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         val project = this.fileBuffer[request.info.projectId]
         if (project == null) {
             val errorStatus = Status.INVALID_ARGUMENT.withDescription(
-                    "There is no project in buffer with project id ${request.info.projectId}")
+                    "There is no project in buffer with project=${request.info.projectId}")
             LOG.error(errorStatus.description)
             responseObserver.onError(StatusException(errorStatus))
             return
@@ -241,7 +241,7 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         val response = CosmasProto.GetVersionResponse.newBuilder()
         if (blob == null) {
             val errorStatus = Status.NOT_FOUND.withDescription(
-                    "There is no such file or file version in storage")
+                    "There is no such file=${request.fileId} or file version in storage")
             LOG.error(errorStatus.description)
             responseObserver.onError(StatusException(errorStatus))
             return
@@ -264,9 +264,9 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         }
         if (versionList.isEmpty()) {
             val description = if (request.startGeneration != -1L) {
-                "There is no file in storage with file id ${request.fileId}"
+                "There is no file=${request.fileId} in storage"
             } else {
-                "There is no versions of file with id ${request.fileId} before generation ${request.startGeneration}"
+                "There is no versions of file=${request.fileId} before generation=${request.startGeneration}"
             }
             val errorStatus = Status.NOT_FOUND.withDescription(description)
             LOG.error(description)
--- a/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
+++ b/src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
@@ -278,8 +278,12 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
     }
 
     private fun getFileVersionListFromMemory(projectInfo: ProjectInfo, fileId: String): List<FileVersionInfo> {
-        return this.fileBuffer[projectInfo.projectId]?.get(fileId)?.historyWindowList
-                ?: emptyList()
+        val project = synchronized(this.fileBuffer) {
+            this.fileBuffer.getOrPut(projectInfo.projectId) { ConcurrentHashMap() }
+        }
+        val fileVersion = this.fileBuffer[projectInfo.projectId]?.get(fileId)
+                ?: restoreFileFromStorage(fileId, projectInfo, project)
+        return fileVersion.historyWindowList
     }
 
     private fun getFileVersionListFromStorage(projectInfo: ProjectInfo, fileId: String,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -481,6 +461,42 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         responseObserver.onCompleted()
     }
 
+    fun restoreFileFromStorage(fileId: String, projectInfo: ProjectInfo,
```
dbarashev: could be private

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -481,6 +461,42 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
...
+                               project: ConcurrentMap<String, CosmasProto.FileVersion>): FileVersion {
+        synchronized(project) {
+            // Getting last version from storage or default instance if it doesn't exist
+            val latestVersionBlob: Blob? = this.storage.get(getBlobId(fileId, projectInfo))
+            val latestVersion = if (latestVersionBlob != null) {
+                CosmasProto.FileVersion.parseFrom(latestVersionBlob.getContent())
+            } else {
+                val bufferVersion = FileVersion.getDefaultInstance()
+                project[fileId] = bufferVersion
```
dbarashev: Can we avoid modifying project here? It seems that the returned result will be written to project map anyway in the calling method, while keeping this function read-only will make code more simple and robust.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -481,6 +461,42 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
         responseObserver.onCompleted()
     }
 
+    fun restoreFileFromStorage(fileId: String, projectInfo: ProjectInfo,
+                               project: ConcurrentMap<String, CosmasProto.FileVersion>): FileVersion {
+        synchronized(project) {
+            // Getting last version from storage or default instance if it doesn't exist
+            val latestVersionBlob: Blob? = this.storage.get(getBlobId(fileId, projectInfo))
+            val latestVersion = if (latestVersionBlob != null) {
```
dbarashev: This combination of `if` with `return` looks uncomfortable. I suggest appending null-case to the previous line with Elvis-operator:

```
            val latestVersionBlob: Blob = this.storage.get(getBlobId(fileId, projectInfo)) ?: return FileVersion.getDefaultInstance() 
            val latestVersion = CosmasProto.FileVersion.parseFrom(latestVersionBlob.getContent())
```

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: ok, done

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -264,9 +263,13 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
             return
         }
         if (versionList.isEmpty()) {
-            val errorStatus = Status.NOT_FOUND.withDescription(
-                    "There is no file in storage with file id ${request.fileId}")
-            LOG.error(errorStatus.description)
+            val description = if (request.startGeneration != -1L) {
+                "There is no file=${request.fileId} in storage"
+            } else {
+                "There is no versions of file=${request.fileId} before generation=${request.startGeneration}"
```
dbarashev, line 269: will it always print "before generation=-1" ?

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -116,6 +117,39 @@ class CosmasGoogleCloudServiceTest {
         assertEquals("kek lol", file)
     }
 
+    @Test
+    fun restoreToBufferCreatePatch() {
+        val fakeStorage: Storage = mock(Storage::class.java)
+        val blob1 = getMockedBlob("ver1", 0)
+        val blob2 = getMockedBlob("ver2", 1)
+        Mockito.`when`(fakeStorage.get(eq(service.getBlobId(FILE_ID, projectInfo(), null))))
+                .thenReturn(blob1).thenReturn(blob2)
```
dbarashev: I suggest moving the second `thenReturn` lower after call to `addPatchToService` to make it clear that we return blob2 only after that

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -116,6 +117,39 @@ class CosmasGoogleCloudServiceTest {
...
+
+    @Test
+    fun restoreToBufferVersionList() {
+        val fakeStorage: Storage = mock(Storage::class.java)
+        val blob1 = getMockedBlob("ver1", 0)
+        val blob2 = getMockedBlob("ver2", 1)
+        Mockito.`when`(fakeStorage.get(eq(service.getBlobId(FILE_ID, projectInfo(), null))))
+                .thenReturn(blob1).thenReturn(blob2)
+        this.service = CosmasGoogleCloudService(this.BUCKET_NAME, fakeStorage, getMockedClock())
+        assertEquals(listOf(0L), getVersionsList(FILE_ID, PROJECT_ID))
```
dbarashev, line 149: Why there is only 1 value in the list, despite that we seem to be operating with two blobs above?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: it should be `request.startGeneration == -1L`, fixed

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: ok

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: It was pretty strange test, fixed

dbarashev (approved):
LGTM

PR #63, opened 2019-02-25
Checking that all files in versions list exist

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -297,7 +297,17 @@ class CosmasGoogleCloudService(private val freeBucketName: String,
             responseObserver.onError(StatusException(errorStatus))
             return@logging
         }
-        response.addAllVersions(versionList)
+        val actualVersionList = mutableListOf<FileVersionInfo>()
+        for (version in versionList) {
+            // checking that files from window really exist - storage.get doesn't load all file content
+            val blob = this.storage.get(getBlobId(version.fileId, request.info, version.generation))
```
dbarashev: I thought we discussed fetching only those objects which are likely to be dead because of timestamp comparison, no?

iisuslik43: yep, done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -1099,6 +1108,49 @@ class CosmasGoogleCloudServiceTest {
             addPatchToService(patch)
             commit()
         }
+        return fakeStorage
+    }
+
+    @Test
+    fun versionsNotExistFromMemory() {
+        val month = 24 * 60 * 60 * 1000
```
dbarashev: is it month really? o_0

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -1099,6 +1108,49 @@ class CosmasGoogleCloudServiceTest {
...
+        val clock = mock(Clock::class.java)
+        Mockito.`when`(clock.millis())
+                .thenReturn(0L)
+                .thenReturn(1L)
+                .thenReturn(month + 1L)
+                .thenReturn(month + 2L)
+                .thenReturn(0L)
```
dbarashev: Break these thenReturns into two blocks? If I get it right, first few values are needed for addManyVersions, while last few are used in getVersionsList ? In this case your can move the tail after addManyVersions call

src/main/resources/logback.xml
```diff
@@ -1,5 +1,5 @@
 
-<configuration debug="true" scan="true">
+<configuration debug="false">
```
dbarashev: I suggest leaving scan=true because otherwise we just have no way to switch on debug level in the runtime

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: ups, fixed

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: first 4 values are used in `addManyVersions`, next 4 in commit inside `addManyVersions`, and only the last one is used in `getVersionsList`, so I splitted it like this

src/main/resources/logback.xml
iisuslik43: done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -1113,43 +1113,51 @@ class CosmasGoogleCloudServiceTest {
...
         val fakeStorage = addManyVersions(4, 5, clock)
         Mockito.`when`(fakeStorage.get(eq(service.getBlobId(FILE_ID, projectInfo(), 2))))
                 .thenReturn(null)
         Mockito.`when`(fakeStorage.get(eq(service.getBlobId(FILE_ID, projectInfo(), 1))))
                 .thenReturn(null)
+
+        Mockito.`when`(clock.millis())
+                .thenReturn(day + 3L)
+
         assertEquals(listOf(4L, 3L), getVersionsList(FILE_ID, PROJECT_ID))
```
dbarashev, line 1143: Please write a comment explaining this expectation. Like 
"We created a sequence of versions with timestamps [day,...., day+2] and version ttl is 2 days. When we ask version list at day+3 moment, it should return everything in the interval [.....]"

dbarashev (approved):
LGTM [1 comment]

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: done

Issue #68, opened 2019-03-06
Write rate limit exceeded when deleting many files


dbarashev:
Apparently, when we get a lot of delete file requests in a row (this may happen e.g. when usr deletes a directory), we send the same number of writes to the cemetery which makes GCS unhappy. We need to batch cemetery writes and/or use rate-limited way of sending write operations. In the former case we probably need some way to send a batch of delete requests to Cosmas. Currently we're sending delete requests to Cosmas one-by-one despite that they come as  a batch in Papeeria.

```
06:32:48.204 [grpc-default-executor-871] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:48.311 [grpc-default-executor-866] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:48.345 [grpc-default-executor-889] [INFO] [CosmasGoogleCloudService] <<< forcedFileCommit 
06:32:48.517 [grpc-default-executor-859] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:48.533 [grpc-default-executor-901] [INFO] [CosmasGoogleCloudService] <<< forcedFileCommit 
06:32:48.820 [grpc-default-executor-874] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:49.217 [grpc-default-executor-869] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:49.353 [grpc-default-executor-876] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:49.962 [grpc-default-executor-865] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:50.296 [grpc-default-executor-860] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:50.307 [grpc-default-executor-860] [INFO] [CosmasGoogleCloudService] >>> forcedFileCommit [projectId=XXXXXX, fileId=YYYYYYY]
06:32:50.655 [grpc-default-executor-861] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:50.715 [grpc-default-executor-877] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:51.320 [grpc-default-executor-864] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
06:32:51.415 [grpc-default-executor-867] [INFO] [CosmasGoogleCloudService] <<< deleteFile 
[...]

PR #69, opened 2019-03-11
Delete list of files

iisuslik43:
#68

### DIFF
--- a/src/main/proto/cosmas.proto
+++ b/src/main/proto/cosmas.proto
@@ -117,15 +117,18 @@ message DeletePatchResponse {
     bytes content = 1;
 }
 
-message DeleteFileRequest {
+message DeleteFilesRequest {
     ProjectInfo info = 1;
-    string fileId = 2;
-    string fileName = 3;
-    int64 removalTimestamp = 4;
+    repeated DeletedFileInfo files = 2;
+    int64 removalTimestamp = 3;
 }
 
+message DeletedFileInfo {
+    string fileId = 1;
+    string fileName = 2;
+}
 
-message DeleteFileResponse {
+message DeleteFilesResponse {
 }
 
 message DeletedFileListRequest {
@@ -173,7 +176,7 @@ service Cosmas {
     }
     rpc DeletePatch (DeletePatchRequest) returns (DeletePatchResponse) {
     }
-    rpc DeleteFile (DeleteFileRequest) returns (DeleteFileResponse) {
+    rpc DeleteFiles (DeleteFilesRequest) returns (DeleteFilesResponse) {
     }
     rpc DeletedFileList (DeletedFileListRequest) returns (DeletedFileListResponse) {
     }
--- a/src/main/proto/cosmas.proto
+++ b/src/main/proto/cosmas.proto
@@ -131,6 +131,17 @@ message DeletedFileInfo {
 message DeleteFilesResponse {
 }
 
+message DeleteFileRequest { // Outdated
+    ProjectInfo info = 1;
+    string fileId = 2;
+    string fileName = 3;
+    int64 removalTimestamp = 4;
+}
+
+
+message DeleteFileResponse { // Outdated
+}
+
 message DeletedFileListRequest {
     ProjectInfo info = 1;
 }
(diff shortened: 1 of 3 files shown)


src/main/proto/cosmas.proto
```diff
@@ -117,15 +117,18 @@ message DeletePatchResponse {
     bytes content = 1;
 }
 
-message DeleteFileRequest {
+message DeleteFilesRequest {
```
dbarashev, line 120: Let's pretend that we're going to update production microservice (which is kinda true)! 
If we want to replace a method in the API with newer one, we need to keep old method for a while, so that older clients could use older method while newer client use the new version. In our reality "older clients" is our current Papeeria and newer is one which will be rolled out after these changes.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -1300,15 +1300,17 @@ class CosmasGoogleCloudServiceTest {
     }
 
     private fun deleteFile(fileId: String, fileName: String, time: Long,
-                           info: ProjectInfo = projectInfo()): StreamRecorder<DeleteFileResponse> {
-        val deleteFileRecorder: StreamRecorder<DeleteFileResponse> = StreamRecorder.create()
-        val deleteFileRequest = DeleteFileRequest.newBuilder()
-                .setInfo(info)
+                           info: ProjectInfo = projectInfo()): StreamRecorder<DeleteFilesResponse> {
```
dbarashev, line 1368: Pls add a couple of tests for batch delete.

src/main/proto/cosmas.proto
iisuslik43: done

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
iisuslik43: done

src/main/proto/cosmas.proto
```diff
@@ -117,15 +117,29 @@ message DeletePatchResponse {
...
+
+message DeletedFileInfo {
+    string fileId = 1;
+    string fileName = 2;
+}
+
+message DeleteFilesResponse {
+}
+
+message DeleteFileRequest { // Outdated
```
dbarashev: "Deprecated" is more commonly used term in such cases. There is even special [option in proto language](https://developers.google.com/protocol-buffers/docs/proto#options) to mark deprecated fields: `[deprecated=true]` . Not sure if it can be used here though.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -660,6 +660,41 @@ class CosmasGoogleCloudServiceTest {
...
+                .setFileId("1")
+                .setFileName("kek1")
+                .setRemovalTimestamp(time)
+                .build()
+        val newTomb2 = CosmasProto.FileTomb.newBuilder()
+                .setFileId("2")
+                .setFileName("kek2")
+                .setRemovalTimestamp(time)
+                .build()
+        Mockito.verify(fakeStorage).create(eq(service.getBlobInfo(cemeteryName, projectInfo())),
```
dbarashev, line 694: Any sense to verify that ther were no more interactions with fakeStorage? (because this is the point of batching)

iisuslik43: ok

src/main/proto/cosmas.proto
iisuslik43: ok, `option deprecated = true;` works - it makes methods and classes `@Deprecated` in Java

dbarashev (approved):
LGTM

PR #72, opened 2019-04-05
Changing fileId inside RestoreDeletedFile

iisuslik43:
https://github.com/papeeria/papeeria/issues/1334

### DIFF
--- a/src/main/proto/cosmas.proto
+++ b/src/main/proto/cosmas.proto
@@ -164,8 +164,8 @@ message ForcedFileCommitResponse {
 
 message RestoreDeletedFileRequest {
     ProjectInfo info = 1;
-    string fileId = 2;    // for compatibility I can't rename it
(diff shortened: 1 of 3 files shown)


src/main/proto/cosmas.proto
```diff
@@ -164,7 +164,8 @@ message ForcedFileCommitResponse {
 
 message RestoreDeletedFileRequest {
     ProjectInfo info = 1;
-    string fileId = 2;
+    string fileId = 2;    // for compatibility I can't rename it
```
dbarashev: I thought it wass number that matters, not the name. 
https://stackoverflow.com/questions/45431685/protocol-buffer-does-changing-field-name-break-the-message

dbarashev (approved):
LGTM [1 comment]

src/main/proto/cosmas.proto
iisuslik43: ok, I ll test it

PR #73, opened 2019-04-08
Saving file buffer to GRPC

iisuslik43:


dbarashev:
As we discussed, let's replace fileBuffer with `Cache` + `CacheLoader` and record file id changes in a separate map. This way we can access file from file buffer by its current id, and load FileVersion from GCS and/or load by its old id if necessary.

iisuslik43:
I ve created another PR, so I m closing this one

PR #74, opened 2019-04-08
Cache loader and fileId map

iisuslik43:


build.gradle
```diff
@@ -54,6 +54,7 @@ dependencies {
     compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.3.+'
     compile group: 'ch.qos.logback', name: 'logback-core', version: '1.3.+'
     compile group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.+'
+    compile group: 'com.google.guava', name: 'guava', version: '11.0.2'
```
dbarashev: Guava 11 o_0? 
It is some [seven years old](https://mvnrepository.com/artifact/com.google.guava/guava) and is going to the elementary school this year!

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -643,18 +674,28 @@ class CosmasGoogleCloudService(
...
+                    project.invalidate(change.oldFileId)
                 }
             }
         }
+        this.storage.create(
+                getBlobInfo("${info.projectId}-fileIdMap", info),
+                CosmasProto.FileIdMap.newBuilder().putAllPrevIds(prevIds).build().toByteArray())
+    }
+
+    private fun getPrevIds(projectId: String): Map<String, String> {
```
dbarashev, line 705: It makes sense to cache this mapping in the memory, at least temporarily. Otherwise you will do additional request on e.g. every forced commit even if you don't need to.

dbarashev, line 705: Also, what about giving this function, protocol buffer and GCS object more descriptive name/suffix? E.g. "fileIdChangeMap" to all.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -643,18 +674,28 @@ class CosmasGoogleCloudService(
...
                     val oldVersion = project[change.oldFileId]
-                            ?: restoreFileFromStorage(change.oldFileId, info, project)
-                    project[change.newFileId] = oldVersion
-                    project.remove(change.oldFileId)
+                    project.put(change.newFileId, oldVersion)
+                    project.invalidate(change.oldFileId)
                 }
             }
         }
+        this.storage.create(
```
dbarashev, line 700: Do you want to do it unconditionally, even if `project == null` ?

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I ll do it generalized in next PR

build.gradle
iisuslik43: maven tricked me...

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I think it's important, that this map stores "previous" ids for files, so this name is ok, adding word "map" to it is useless

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I think this `if` is useless, cause project will never be null

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
dbarashev: In my opinion, "file id change map" just can be used consistently in all names:
`getFileIdChangeMap`, `message FileIdChangeMap`, `$projectId-fileIdChangeMap`. Now it is `getPrevIds` vs "fileIdMap".

I don't feel strongly about this, though.

dbarashev (approved):
LGTM
I think we need to remove entries from the file id mapping. The moment after writing a version of a file seems to be okay for that.

PR #77, opened 2019-04-12
Storing fileId in FileVersion

iisuslik43:


src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
```diff
@@ -603,10 +604,12 @@ class CosmasGoogleCloudService(
             latestVersion.patchesList.last().userName
         }
 
+        val latestFileId = if (latestVersion.fileId.isBlank()) fileId else latestVersion.fileId
```
dbarashev, line 614:
```suggestion
        // In most cases latestVersion.fileId == fileId
        // However, when file changes its id (e.g. because of Dropbox import), we can
        // read its latest blob thanks to fileIdChangeMap, and latestVersion from this blob
        // will have different fileId, so we should add to latestVersion's window FileVersionInfo
        // with fileId of latestVersion.
        // In reality this makes the latest version unavailable, while previous versions
        // are still available.
        val latestFileId = if (latestVersion.fileId.isBlank()) fileId else latestVersion.fileId
```

dbarashev (approved):
I added a comment explaining the logic with latestFileId. Please check if my understanding is correct. Also, it seems that we have no unit test covering this edge case, do we? Changes in the test code look pretty technical.

src/main/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudService.kt
iisuslik43: I changed your comment a bit. About testing, `hardRestoring` checks that fileVersionList is correct for this case

PR #82, opened 2019-06-04
Some measures to prevent RAM/CPU over-utilisation

dbarashev:
* Added expiration of per-project caches (meaningful change). They woulld never expire and I suspect that eventually we run out of memory and GC eats all CPU
* Just a single retry to GCS (with 5 retries and exponential backoff we will run out of client RPC deadline anyway)
* Updated tests.

Changes in `CosmasGoogleCloudService` are mostly cosmetical. The most meaningful are: cache expiration and invalidation and  `Ticker` instead of Clock and currentTimeMillis.

The major changes in `CosmasGoogleCloudService` are related to tests which used `addManyVersions` method. They need their custom `service` instance (and that custom instance used to be assigned to `this.service` inside addManyVersions body) and for the purpose of using different CosmasGoogleCloudService instances most of the methods were refactored to accept `cosmas: CosmasGoogleCloudService` argument instead of using `this.service`.
Also, the body of `addManyVersions` was split into mock object building part and actual service calls (and some comments were added). Manipulations with SequenceTicker allow for assigning proper timestamps to mocks (this usually comes first) and then using the same sequence of timestamps in RPC calls which are supposed to operate with mocks.


dbarashev:
@iisuslik43 if you care, feel free to take a look at this too

dbarashev:
Travis CI is currently failing because of build file structure which was adjusted for multi-project build in Papeeria repo. I don't know yet a proper way of fixing this. However, uncommenting versions in `plugins` stanza of `build.gradle` makes this working without multi-project build.

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
```diff
@@ -1678,20 +1711,20 @@ class CosmasGoogleCloudServiceTest {
     }
 
     private fun getVersionsList(fileId: String = FILE_ID, projectId: String = PROJECT_ID,
-                                startGeneration: Long = -1L, isFreePlan: Boolean = true): List<Long> {
+                                startGeneration: Long = -1L, isFreePlan: Boolean = true, cosmas: CosmasGoogleCloudService = this.service): List<Long> {
```
edgarzhavoronkov: This line looks too long, i think(so 1723)

edgarzhavoronkov (approved):
LGTM. There are a couple of long looking lines, but it's minor I think

src/test/kotlin/com/bardsoftware/papeeria/backend/cosmas/CosmasGoogleCloudServiceTest.kt
dbarashev: fixed

PR #83, opened 2019-06-05
Monitoring and bugfix

dbarashev:
* Metricz will be published on port 9700 by default
* Ticker bugfix (it would mark all versions circa 1970 otherwise)


edgarzhavoronkov (approved):
LGTM, my bad, didn't know about Guava's timer

